PR #1, opened 2016-06-20
Unauthorized access to Booking Request API

benlovell:
Trello: https://trello.com/c/oYcaMGdB

`POST /api/v1/booking_requests`

This sets up the namespacing, end point and authorization for the Booking
Requests API. At the moment, to satisfy this particular story I've only added
the bare minimum to `User` to fulfil the GDS-SSO protocol's needs and it isn't
persisted yet.


.rubocop.yml
```diff
@@ -20,3 +20,8 @@ Rails:
 
 Style/FrozenStringLiteralComment:
   Enabled: false
+
+Style/PredicateName:
+  Exclude:
+    # for `has_permission?`
+    - 'app/models/user.rb'
```
benlovell: Yeah, it should. Not sure what I was thinking here.

spec/requests/unauthorized_api_access_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+RSpec.describe 'POST /api/v1/booking_requests' do
+  scenario 'unauthorized access' do
+    given_a_permitted_gds_sso_user
```
dwhenry: I'm a little confused by your intention here, what has this user been "permitted to" given an unauthorised request is then made.

benlovell: This would ultimately be a `User` that has a specific role, to permit API access. I agree the language isn't particularly clear though. I'll change it.

dwhenry:
🚢

PR #2, opened 2016-06-24
Create a valid Booking Request

benlovell:
Completes the _Create a valid Booking Request_ scenario from the card
referenced here: https://trello.com/c/oYcaMGdB

Probably best reviewed commit-by-commit.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -20,3 +20,7 @@ Rails:
 
 Style/FrozenStringLiteralComment:
   Enabled: false
+
+Metrics/MethodLength:
+  Exclude:
+    - 'spec/requests/*'
--- a/app/views/booking_requests/customer.html.erb
+++ b/app/views/booking_requests/customer.html.erb
@@ -1,0 +1,39 @@
+<p style="color: #0B0C0C;font-family: Helvetica, Arial, sans-serif;margin: 15px
+0;font-size: 16px;line-height: 1.315789474;">Dear <%= @booking_request.name %>,</p>
+
+<p style="color: #0B0C0C;font-family: Helvetica, Arial, sans-serif;margin: 15px
+0;font-size: 16px;line-height: 1.315789474;">Thank you for requesting a
+guidance appointment with Pension Wise, we'll confirm the appointment slot
+within the next two working days.</p>
+
+<p style="color: #0B0C0C;font-family: Helvetica, Arial, sans-serif;margin: 15px
+0;font-size: 16px;line-height: 1.315789474;">Your first choice is<br><strong
+   class="emphasize" style="font-size: 24px;"><%= @booking_request.primary_slot %></strong></p>
+
+<p style="color: #0B0C0C;font-family: Helvetica, Arial, sans-serif;margin: 15px
+0;font-size: 16px;line-height: 1.315789474;">Your second choice is<br><strong
+  class="emphasize" style="font-size: 24px;"><%= @booking_request.secondary_slot %></strong></p>
+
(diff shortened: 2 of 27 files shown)


app/views/booking_requests/customer.html.erb
```diff
@@ -0,0 +1,39 @@
+<p style="color: #0B0C0C;font-family: Helvetica, Arial, sans-serif;margin: 15px
```
dwhenry: are we going to support text only emails?

app/views/booking_requests/customer.html.erb
```diff
@@ -0,0 +1,39 @@
...
+guidance appointment with Pension Wise, we'll confirm the appointment slot
+within the next two working days.</p>
+
+<p style="color: #0B0C0C;font-family: Helvetica, Arial, sans-serif;margin: 15px
+0;font-size: 16px;line-height: 1.315789474;">Your first choice is<br><strong
+   class="emphasize" style="font-size: 24px;"><%= @booking_request.primary_slot %></strong></p>
+
+<p style="color: #0B0C0C;font-family: Helvetica, Arial, sans-serif;margin: 15px
+0;font-size: 16px;line-height: 1.315789474;">Your second choice is<br><strong
+  class="emphasize" style="font-size: 24px;"><%= @booking_request.secondary_slot %></strong></p>
```
dwhenry: how does this look if the user has only selected one slot?

benlovell: They will have to select 3. The validation is coming in the next story.

app/views/booking_requests/customer.html.erb
benlovell: We will do. But for now I'm waiting for the text-only equivalent. I mentioned that in the commit.

.rubocop.yml
```diff
@@ -20,3 +20,7 @@ Rails:
 
 Style/FrozenStringLiteralComment:
   Enabled: false
+
+Metrics/MethodLength:
+  Exclude:
+    - 'spec/requests/*'
```
benlovell: Yeah, me too. I know we are going to get a bunch of violations regarding this rule in the request specs by their very nature. Happy to explicitly add each individual request spec, rather than glob them all, if you prefer that?

app/views/booking_requests/customer.html.erb
dwhenry: I'll leave it up to you. Not sure any real value in listing each file, especially as chances are you will need to list ever file.

db/schema.rb
```diff
@@ -11,11 +11,35 @@
...
+    t.string   "age_range",                  null: false
+    t.boolean  "accessibility_requirements", null: false
+    t.boolean  "marketing_opt_in",           null: false
+    t.boolean  "defined_contribution_pot",   null: false
+    t.datetime "created_at",                 null: false
+    t.datetime "updated_at",                 null: false
+  end
+
+  create_table "slots", force: :cascade do |t|
+    t.integer  "booking_request_id"
```
dwhenry: should we index this field?

.rubocop.yml
dwhenry: I'll leave it up to you. Not sure any real value in listing each file, especially as chances are you will need to list ever file.

benlovell: That's my thoughts too. I'll leave as-is.

db/schema.rb
benlovell: Oh, that's weird. I was sure the `belongs_to` in a migration adds that already. I'll sort it.

dwhenry:
image

dwhenry:
image

PR #3, opened 2016-06-25
Invalid Booking Request

benlovell:
Serializes validation errors (using Rails built-in serialization) in a
simple fashion and returns an HTTP 422 response code.

### DIFF
--- a/app/controllers/api/v1/booking_requests_controller.rb
+++ b/app/controllers/api/v1/booking_requests_controller.rb
@@ -6,13 +6,21 @@ module Api
       before_action :authorise_api_user!
 
       def create
-        booking_request = BookingRequest.create!(booking_request_params)
-        BookingRequests.customer(booking_request).deliver_later
+        booking_request = BookingRequest.new(booking_request_params)
 
-        head :created
+        if booking_request.save
+          BookingRequests.customer(booking_request).deliver_later
+          head :created
+        else
+          render_errors(booking_request)
+        end
       end
 
       private
+
+      def render_errors(booking_request)
+        render json: { errors: booking_request.errors }, status: :unprocessable_entity
+      end
 
       def booking_request_params # rubocop:disable Metrics/MethodLength
         params.require(:booking_request).permit(
--- a/app/models/slot.rb
+++ b/app/models/slot.rb
@@ -1,5 +1,14 @@
 class Slot < ActiveRecord::Base
+  PERMITTED_TIME_REGEX = /\A\d{4}\z/
+  PERMITTED_PRIORITIES = [1, 2, 3].freeze
+
   belongs_to :booking_request
+
+  validates :date, presence: true
+  validates :priority, inclusion: { in: PERMITTED_PRIORITIES }
+  validates :from, format: { with: PERMITTED_TIME_REGEX }
+  validates :to, format: { with: PERMITTED_TIME_REGEX }
+  validate :validate_from_before_to
 
   def morning?
     from < '1300'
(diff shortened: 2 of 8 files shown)


app/controllers/api/v1/booking_requests_controller.rb
```diff
@@ -6,14 +6,22 @@ class BookingRequestsController < ActionController::Base
...
+          head :created
+        else
+          render_errors(booking_request)
+        end
       end
 
       private
 
+      def render_errors(booking_request)
+        render json: { errors: booking_request.errors }, status: :unprocessable_entity
```
dwhenry: Does this correctly format slot validations or do we just get `slot is invalid`?

Does this group the `slots` priority validation from the `booking request` object with the validations from the `slot` object?

Do we care (i.e. are we planning to use these validation messages anywhere)?

app/models/slot.rb
```diff
@@ -12,4 +21,10 @@ def to_s
   def period
     morning? ? 'morning' : 'afternoon'
   end
+
+  private
+
+  def validate_from_before_to
+    errors.add(:base, 'from must be before to') unless from < to
```
dwhenry:
```
validates :from, numericality: { less_than: ->(booking_request) { booking_request.to }, message: 'must be before to' }
```

spec/requests/create_a_booking_request_spec.rb
```diff
@@ -86,4 +93,32 @@ def and_the_service_responds_with_a_201
...
+  end
+
+  def then_the_service_responds_with_a_422
+    expect(response).to be_unprocessable
+  end
+
+  def and_the_errors_are_serialized_as_json
+    json = JSON.parse(response.body)
+
+    expect(json['errors']['name']).to be
```
dwhenry: I'm not a fan of the hanging `be`. I would like to see it explicitly check for something
```
expect(json['errors']['name']).to be_present
```

app/controllers/api/v1/booking_requests_controller.rb
benlovell: >Does this correctly format slot validations or do we just get slot is invalid?

>Does this group the slots priority validation from the booking request object with the validations from the slot object?

Yes. The output looks something like this:

```javascript
{
  "errors": {
    "name": [
      "can't be blank"
    ],
    "email": [
      "can't be blank"
    ],
    "phone": [
      "can't be blank"
    ],
    "memorable_word": [
      "can't be blank"
    ],
    "age_range": [
      "is not included in the list"
    ],
    "accessibility_requirements": [
      "is not included in the list"
    ],
    "marketing_opt_in": [
      "is not included in the list"
    ],
    "defined_contribution_pot": [
      "is not included in the list"
    ],
    "slots": [
      "you must provide a slot for each permitted priority (1, 2, 3)"
    ]
  }
}
```

>Do we care (i.e. are we planning to use these validation messages anywhere)?

At this stage, no. We'll log them to diagnose any issues that may occur, but the likelihood of requests getting this far is entirely exceptional circumstance, and would highlight an issue in the client. There will be a time when locations are supplanted by something over here - likely when booking centres need to manage their own slots and availability - at which time this will feel somewhat more cohesive.

app/models/slot.rb
benlovell: I'm gonna say I prefer my version to be honest...

spec/requests/create_a_booking_request_spec.rb
benlovell: This verifies that `errors` are present, with a child `name`. Knowing how errors are structured, this gives me enough confidence to know things are being serialized / deserialized as I expect. `name` would not be present in the output unless it has validation errors, so I'm not sure what (other than personal preference, of course) adding a presence check buys us in this example?

I don't feel too strongly about this though, so if you do I'm happy to change.

app/controllers/api/v1/booking_requests_controller.rb
dwhenry: I'm happy with that.

dwhenry:
🚢  when ready

PR #5, opened 2016-07-12
Include twilio booking number in confirmation

benlovell:
Includes the booking location specific contact number when emailing the
customer with their booking request confirmation.

I moved this into a background job that runs inline for the time being.
The thinking here was that both the locations API call and email
delivery are better done out-of-band. It felt wrong to make the
locations API call in the mailer itself, too.

### DIFF
--- a/app/views/booking_requests/customer.html.erb
+++ b/app/views/booking_requests/customer.html.erb
@@ -35,5 +35,5 @@ class="emphasize" style="font-size: 24px;"><%=
 
 <p style="color: #0B0C0C;font-family: Helvetica, Arial, sans-serif;margin: 15px
 0;font-size: 16px;line-height: 1.315789474;">To cancel or change your
-appointment, or if any of the information above isn't correct, call 0800 138
-3944 or reply to this email.</p>
+appointment, or if any of the information above isn't correct, call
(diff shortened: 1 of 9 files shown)


benbarnett:
Nice, ⚓ ⬆️

app/views/booking_requests/customer.html.erb
```diff
@@ -35,5 +35,5 @@ class="emphasize" style="font-size: 24px;"><%=
 
 <p style="color: #0B0C0C;font-family: Helvetica, Arial, sans-serif;margin: 15px
 0;font-size: 16px;line-height: 1.315789474;">To cancel or change your
-appointment, or if any of the information above isn't correct, call 0800 138
-3944 or reply to this email.</p>
+appointment, or if any of the information above isn't correct, call
+ <%= @booking_location.online_booking_twilio_number %> or reply to this email.</p>
```
dwhenry: Do we want to use the Twilio number here?

As this could distort the stats the Twilio number is designed to capture i.e. number of appointments coming from the website via phone channel.

benlovell: These are pilot-specific numbers created specifically for this purpose.

PR #10, opened 2016-07-14
Store booking location ID from API payload

benlovell:
This means we don't have to perform expensive lookups for the booking
location sub-tree any time we query booking requests for a specific
booking's manager.


benbarnett:
Really lovely

PR #11, opened 2016-07-14
Bind booking requests

benlovell:
Displays booking requests for the context booking manager.

The only complexity is the decorated booking request, this is to add the
location awareness and save an API call roundtrip per location bound to the
UI.

For now the fulfilment button is stubbed but this will be addressed in the
upcoming stories.


app/models/location_aware_booking_requests.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def each
+    return enum_for(:each) unless block_given?
+
+    __getobj__.each do |booking_request|
+      yield LocationAwareBookingRequest.new(
+        booking_location: booking_location,
+        booking_request: booking_request
+      )
+    end
```
dwhenry: The more I look at this the more I think it is more complex than needed.. 

We have turned an array of objects into a first class object then used a delegator to modify the each behaviour. 

Why not just map the array that way you don't need the SimpleDelegator?

app/models/user.rb
```diff
@@ -1,8 +1,14 @@
...
 class User < ActiveRecord::Base
   PENSION_WISE_API_PERMISSION = 'pension_wise_api_user'
+  BOOKING_MANAGER_PERMISSION  = 'booking_manager'
 
   include GDS::SSO::User
 
   serialize :permissions, Array
+
+  has_many :booking_requests,
+           -> { includes(:slots).order(:created_at) },
```
dwhenry: can we change this to:

```
-> { order(:created_at).includes(:slots) },
```

doesn't effect the functionality but my mind spent 20min trying to work out why you where ordering slots by created at.. :(

app/models/location_aware_booking_requests.rb
benlovell: Because it needs to answer to `#empty?` and other enumerable behaviour - is the short explanation. I could just delegate `#empty?` and `#each` without deriving from `SimpleDelegator` sure, but that's just another approach with the same result and I'm not sure I see the benefits either way.

What do you think about this:

```ruby
class LocationAwareBookingRequests
  def initialize(booking_requests, booking_location)
     # blah
  end

  def all
    @all ||= @booking_requests.map { |b| LocationAwareBookingRequest.new(b, @location)
  end
end
```

Changing the interface slightly but achieving the same. Albeit a little simpler.

app/models/user.rb
benlovell: OK I'll do that, just for you 😉

app/models/location_aware_booking_requests.rb
benlovell: I changed this in f40add7. It's definitely simpler.

app/models/user.rb
benlovell: Resolved with 0ff071f.

app/models/user.rb
```diff
@@ -1,8 +1,14 @@
...
+  BOOKING_MANAGER_PERMISSION  = 'booking_manager'
 
   include GDS::SSO::User
 
   serialize :permissions, Array
+
+  has_many :booking_requests,
+           -> { order(:created_at).includes(:slots) },
+           primary_key: :organisation_content_id,
+           foreign_key: :booking_location_id
```
dwhenry: Just wanted to check the thinking around `organisation_content_id` and how it differs from `booking_location_id`.

benlovell: Check the rationale / commentary I just added in e11136b.

dwhenry:
😄  looks good to me. 👕 or 🚢

PR #12, opened 2016-07-15
Bump `booking_locations` for `Accept` header

benlovell:
Ensures API calls are correctly authenticated, using the bearer token.


benbarnett:
👍

PR #15, opened 2016-07-17
Cache bearer tokens for API authentication

benlovell:
Bumps GDS-SSO and configures the in-memory cache for bearer tokens.

For background see: https://github.com/alphagov/gds-sso/pull/97


benbarnett:
👍

PR #16, opened 2016-07-24
Fulfil booking requests

benlovell:
Allows the booking manager to fulfil booking requests. This can be reviewed on a commit-by-commit basis but I'll be squashing it down to something more meaningful for permanent record.


app/forms/appointment_form.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def date
+    @date ||= location_aware_booking_request.primary_slot.date
+  end
+
+  def time
+    @time ||= location_aware_booking_request.primary_slot.delimited_from
+
+    Time.zone.parse("2016-01-01 #{@time}")
```
dwhenry: given the date part is ignored can we not just use 
```
 Time.zone.parse("#{@time}")
```
or does it break with daylight saving?

benlovell: Nope, that just never occurred to me. I'll change it 👍

app/models/slot.rb
```diff
@@ -15,11 +15,19 @@ def morning?
   end
 
   def to_s
-    "#{date.to_s(:gov_uk)} - #{period}"
+    "#{formatted_date} - #{period}"
+  end
+
+  def formatted_date
+    date.to_s(:gov_uk)
```
benlovell: This should be a `govuk_date` instead.

benlovell:
@benbarnett Lovely job.

benlovell:
Squashed that fixup commit.

PR #17, opened 2016-07-26
Mount sidekiq web control panel

benlovell:
This is secured with `SIDEKIQ_WEB_{USERNAME,PASSWORD}`.

The variables are fetched from the environment so if they're not
present the application won't start.


benbarnett:
🚢

PR #18, opened 2016-07-26
Just permit access to any authenticated users

benlovell:
We were previously authenticating via HTTP basic, using the existing
authenticated admin account makes more sense.


benbarnett:
👍

PR #19, opened 2016-07-26
Process the `mailers` queue

benlovell:
Whoops, this queue wasn't being processed so email deliveries were not
happening asynchronously.


benbarnett:
👍

PR #20, opened 2016-07-27
Ensure fulfilled booking requests are hidden

benlovell:
Prior to this, when booking requests transitioned into appointments
they were still being displayed on the booking manager's list of
unfulfilled booking requests. Now we hide them once they're fulfilled
so the booking manager can view the list as things that need immediate
attention.


benbarnett:
👍

PR #23, opened 2016-07-27
Notify booking manager for new booking requests

benlovell:
Sends an email notification alerting the booking managers of a new
booking request.

There is a very slight timing issue, whereby a booking manager may not
have yet signed into the application via signon and thus not actually
exist within the `Users` table. In this case we raise
`BookingManagersNotFoundError` and allow the job to retry indefinitely
using the typical sidekiq exponential back-off strategy.

The actual likelihood of this occurring is incredibly slim as at least
one booking manager per-location is likely to have signed into the
production environment before customers have placed booking requests.

### DIFF
--- a/app/jobs/booking_manager_confirmation_job.rb
+++ b/app/jobs/booking_manager_confirmation_job.rb
@@ -1,0 +1,16 @@
+BookingManagersNotFoundError = StandardError
+
+class BookingManagerConfirmationJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(booking_request)
+    booking_location = BookingLocations.find(booking_request.location_id)
+    booking_managers = User.where(organisation_content_id: booking_location.id)
+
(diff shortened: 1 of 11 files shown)


app/jobs/booking_manager_confirmation_job.rb
```diff
@@ -0,0 +1,10 @@
+class BookingManagerConfirmationJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(booking_request)
+    booking_location = BookingLocations.find(booking_request.location_id)
+    booking_manager  = User.find_by!(organisation_content_id: booking_location.id)
```
benbarnett: Worth finding *all* users with booking manager role for this location?

benlovell: Yes, that is a very good point. 😵 🌴 

Will deal with that now.

benbarnett:
👍 👍

PR #27, opened 2016-07-28
Display a list of appointments

benlovell:
Shows a list of appointments for the booking manager's location. At the
moment this is unfiltered, and ordered by `proceeded_at`.

The appointment status is displayed within the appointment list also.


benbarnett:
💯

app/models/appointment.rb
```diff
@@ -1,4 +1,14 @@
 class Appointment < ActiveRecord::Base
+  enum status: %i(
+    pending
+    completed
+    no_show
+    ineligible_age
+    ineligible_pension_type
+    cancelled_by_customer
+    cancelled_by_pension_wise
+  )
```
dwhenry: do we really want to store status in the DB as an integer with a special meaning that is only defined here? 

can't we just store it as a text field?

 or if we have to use integers use a lookup table.

On a plus side this approach means that if you change the ordering in this list you break historical data. :)

benlovell: tl;dr I'm OK with this as-is.

We won't be integrating directly with the database, so storing the names of the statuses in a denormalised way makes no sense here. Using a lookup table is adding another unnecessary join, and we don't need the lookups to be editable.

With regards ordering in enums - if you change the ordering of an enum then you shouldn't be trusted to write code. Enums are inherently ordered structures. 😄

PR #28, opened 2016-07-28
Guard against multiple fulfilments of a booking

benlovell:
There is a chance that given multiple booking managers for the same
location, when they receive the new booking notification they all race
to fulfil the booking request.

Previously, it was possible that a single booking request be fulfilled
multiple times. This change ensures the booking request is not already
associated with an appointment, thus guarding against this outcome.


benbarnett:
👍

PR #31, opened 2016-07-29
Cache booking locations for 2 hours

benlovell:
Caches booking locations using a read-through cache. This will use
whatever backend Rails caching is configured with - currently
in-memory.

### DIFF
--- a/config/initializers/booking_locations.rb
+++ b/config/initializers/booking_locations.rb
@@ -1,4 +1,8 @@
-unless Rails.env.production? || Rails.env.staging?
+if Rails.env.production? || Rails.env.staging?
+  require_relative '../../lib/cached_booking_locations_api'
+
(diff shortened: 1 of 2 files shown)


config/initializers/booking_locations.rb
```diff
@@ -1,4 +1,8 @@
-unless Rails.env.production? || Rails.env.staging?
+if Rails.env.production? || Rails.env.staging?
+  require_relative '../../lib/cached_booking_locations_api'
```
dwhenry: as this is load in the ruby env it should work with just:
```
require 'cached_booking_locations_api'
```

benlovell: Yeah, I noticed that after the fact. I'll modify this shortly. 😄

PR #32, opened 2016-07-30
Edit Appointments

benlovell:
Lets the bookings manager edit an existing appointment. This allows them to alter the customer details, change the guider, location or date and time of an appointment and also change the status of an appointment once complete or otherwise.

We send the customer a notification email upon successful update of the appointment, so long as the status hasn't changed. As-is status changes are mutually exclusive from other customer or appointment detail updates so we simply send the regular fulfilment email - but with a lead in paragraph explaining the appointment has been updated.

We also audit all changes to the appointment using the `audited` gem. This is pretty basic as it stands now, and we don't plan to reify old versions or provide any of those kind of capabilities in the near-future. For now we just store the simple changeset, without presentation of any kind.

screen shot 2016-08-02 at 14 16 57

### DIFF
--- a/app/forms/edit_appointment_form.rb
+++ b/app/forms/edit_appointment_form.rb
@@ -41,6 +41,17 @@ class EditAppointmentForm
     FlattenedLocationMapper.map(booking_location)
   end
 
+  def update
+    location_aware_appointment.update(
+      name: name,
+      email: email,
+      phone: phone,
+      guider_id: guider_id,
+      location_id: location_id,
+      proceeded_at: proceeded_at
+    )
+  end
+
   # memoized readers to use `params` when present
   EDITABLE_ATTRIBUTES.each do |attribute|
     class_eval <<~EORUBY, __FILE__, __LINE__ + 1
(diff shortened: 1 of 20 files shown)


app/forms/edit_appointment_form.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    Time.zone.parse(@proceeded_at.to_s)
+  end
+
+  def flattened_locations
+    FlattenedLocationMapper.map(booking_location)
+  end
+
+  # memoized readers to use `params` when present
+  EDITABLE_ATTRIBUTES.each do |attribute|
```
benlovell: I'm not sure - given the small set of `EDITABLE_ATTRIBUTES` whether just writing these out in regular code is just a simpler approach. Personally, I have trouble with metaprogramming when it's contained.

app/forms/edit_appointment_form.rb
```diff
@@ -0,0 +1,64 @@
...
+    class_eval <<~EORUBY, __FILE__, __LINE__ + 1
+      def #{attribute}
+        @#{attribute} ||= location_aware_appointment.#{attribute}
+      end
+    EORUBY
+  end
```
dwhenry: why are we memorising attributes? could this not just be:
```
delegate *EDITABLE_ATTRIBUTES, to: :location_aware_appointment
```

benbarnett:
👍 👍 👍

PR #33, opened 2016-08-01
Tag emails with `message_type` in headers

benbarnett:
Allows for more granular analysis of email deliveries and
interaction.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,8 +1,9 @@
+require 'mailgun_headers'
+
 class ApplicationMailer < ActionMailer::Base
-  default(
-    from: 'appointments@pensionwise.gov.uk',
-    'X-Mailgun-Variables' => { online_booking: true }.to_json
-  )
+  include MailgunHeaders
+
+  default from: 'appointments@pensionwise.gov.uk'
 
   layout 'mailer'
 end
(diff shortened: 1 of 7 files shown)


app/mailers/application_mailer.rb
```diff
@@ -1,8 +1,7 @@
+require 'mailgun_utils'
+
 class ApplicationMailer < ActionMailer::Base
-  default(
-    from: 'appointments@pensionwise.gov.uk',
-    'X-Mailgun-Variables' => { online_booking: true }.to_json
-  )
+  default(from: 'appointments@pensionwise.gov.uk')
```
benlovell: Can we remove the parens from this now?

app/mailers/appointments.rb
```diff
@@ -1,10 +1,13 @@
 class Appointments < ApplicationMailer
+  include MailgunUtils
```
benlovell: May as well just include this in `ApplicationMailer` once instead?

lib/mailgun_utils.rb
```diff
@@ -0,0 +1,14 @@
+module MailgunUtils
```
benlovell: Nit-picky but the name feels a bit scuzzy. Can we call this `MailgunHeaders` or something instead?

benbarnett: Yep, good idea.

benlovell:
Lovely. 🚢 that 💩

PR #34, opened 2016-08-01
Strip personal information from logs

benlovell:
Strips any personal information (provided via parameters) from being logged.


benbarnett:
👍

PR #38, opened 2016-08-04
Cached booking locations

benlovell:
Cache booking locations in memory for 2 hours. This wasn't 100% effective prior to these changes.

### DIFF
--- a/lib/cached_booking_locations_api.rb
+++ b/lib/cached_booking_locations_api.rb
@@ -1,15 +1,19 @@
-class CachedBookingLocationsApi < BookingLocations::Api
-  def initialize(cache: Rails.cache)
+class CachedBookingLocationsApi
+  def initialize(cache: Rails.cache, booking_locations_api: BookingLocations::Api.new)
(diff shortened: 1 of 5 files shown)


lib/cached_booking_locations_api.rb
```diff
@@ -1,11 +1,11 @@
-class CachedBookingLocationsApi < BookingLocations::Api
+class CachedBookingLocationsApi
   def initialize(cache: Rails.cache)
     @cache = cache
   end
 
   def find(location_id)
     cache.fetch(location_id, expires_in: 2.hours) do
```
dwhenry: If is worth making the key more specific so it can't conflict if we use other caches later:
```
cache.fetch("Location: #{location_id}", expires_in: 2.hours) do
```

benlovell: It's a uuid, there's no chance of it clashing and besides we're not caching anything else right now.

benbarnett:
Noice 👍

PR #39, opened 2016-08-04
Cache via the booking locations gem

benlovell:
Booking locations now provides this functionality for us, so long as we
configure the cache with `Rails.cache`. Let's use that instead.


benbarnett:
👍

PR #40, opened 2016-08-08
Paginate booking requests list

benlovell:
Paginates the booking requests list with a default page size of 10.
Styled using the typical bootstrap 3 compatible kaminari views.

Rather than start delegating kaminari required things through the
decorated booking requests, I just exposed the underlying (undecorated)
arel and therefore kaminari paginateable thing.

screen shot 2016-08-08 at 11 43 44

### DIFF
--- a/app/views/booking_requests/index.html.erb
+++ b/app/views/booking_requests/index.html.erb
@@ -9,9 +9,11 @@
 
 <div class="row">
   <div class="col-md-12">
-    <% if @booking_requests.empty? %>
+    <% if @booking_requests.all.empty? %>
       <p class="no-content t-notice">No booking requests requiring attention.</p>
     <% else %>
+      <%= paginate @booking_requests.entities %>
+
       <table class="table table-bordered">
         <thead>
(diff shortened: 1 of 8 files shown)


benbarnett:
Neat! 👍

app/views/booking_requests/index.html.erb
```diff
@@ -9,9 +9,11 @@
 
 <div class="row">
   <div class="col-md-12">
-    <% if @booking_requests.empty? %>
+    <% if @booking_requests.all.empty? %>
       <p class="no-content t-notice">No booking requests requiring attention.</p>
     <% else %>
+      <%= paginate @booking_requests.entities %>
```
dwhenry: feels like naming is the wrong way around. 

maybe rename `.all` to `.page`?

benlovell: I agree. This `LocationAwareEntities` decorator is used for displaying appointments too. I'm going to get those paginated then circle back around and tidy up this naming.

PR #41, opened 2016-08-08
Remove minimum permitted date

benlovell:
This was blocking the booking manager from fulfilling the booking
request with an appointment date prior to the customer's first selected
slot choice.


benbarnett:
There's server side validation to fall-back to should Date < today be selected right?

If so 👍

benlovell:
@benbarnett Indeedly https://github.com/guidance-guarantee-programme/planner/blob/597286efe8f63f33073aca9e66cfa33360ced57d/app/forms/appointment_form.rb#L74

PR #42, opened 2016-08-08
Paginate appointments

benlovell:
Paginates appointments the same as booking requests.

Also improve naming for paging: `#all` doesn't make much sense now we're
actually returning a paginated set. This improves the naming and makes it more
obvious what we're dealing with in the front end.


benbarnett:
🤘

PR #43, opened 2016-08-09
Tag emails with environment they were sent from

benbarnett:
As emails are delivered using the production mailgun environment,
this allows us to identify the source and whether to include them
in our dashboards.


benlovell:
:shipit:

PR #44, opened 2016-08-10
Appointment statistics

benlovell:
Calculates and persists statistics for fulfilment time and difference between the customer's first slot and the appointment date / time they were given.

### DIFF
--- a/spec/models/appointment_spec.rb
+++ b/spec/models/appointment_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe Appointment do
         subject.save!
 
         expect(subject.fulfilment_time_seconds).to eq(
-          subject.proceeded_at.to_i - subject.booking_request.created_at.to_i
+          (subject.proceeded_at.to_i - subject.booking_request.created_at.to_i).abs
         )
       end
     end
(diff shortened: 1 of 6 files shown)


spec/models/appointment_spec.rb
```diff
@@ -11,6 +11,42 @@
     expect(subject.reference).to eq(subject.booking_request.reference)
   end
 
+  describe '#fulfilment_time_seconds' do
+    subject { build(:appointment) }
+
+    it 'defaults to 0' do
+      expect(described_class.new.fulfilment_time_seconds).to be_zero
```
dwhenry: I take it you can tell which records have a valid value as they have a `proceeded_at` time rather than using a `nil` default value to determine this?

benlovell: Yeah, in practice that scenario can never happen. If the record has a `proceeded_at` - which incidentally is a user driven thing, not a typical rails timestamp - then it will have entries for these statistics too.

Also to preempt any discussion around this approach - we spoke about pushing a full, serialized version of the `Appointment` and its associated stuff to a queue, and processing these kinds of things out-of-band but that's perhaps something for the future...

dwhenry: :) kool

PR #48, opened 2016-08-16
Corrects stats for fulfilment time

benlovell:
This should look to the difference between the `created_at` timestamps
of both the appointment and underlying booking request.

Since this calculation can be performed when the appointment is
initially persisted - and since Rails (rightly) doesn't default the
`created_at` timestamp we depend on for our calculation - I provided a
default value that serves this purpose. Rails will do the Right Thing
as far as setting this timestamp is concerned, after it is persisted.

Since the calculations are performed against seconds after the epoch
the potential for drift isn't something we should be concerned about.

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -11,7 +11,7 @@ class Appointment < ActiveRecord::Base
     cancelled_by_pension_wise
   )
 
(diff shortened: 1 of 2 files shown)


benbarnett:
👍

benlovell:
Reminder to self: the existing stats need recalculating across all environments.

PR #50, opened 2016-08-17
[DO NOT MERGE] UI for Activity Feed

benbarnett:
Includes some ideas for different types of activity, although
these should be mapped out properly.

Initially, user generated notes should suffice as the booking
managers can use this to say if the customer has been contacted.

screen shot 2016-08-17 at 16 03 23


benlovell:
I think I have what I need from here now. So it's goodbye...

PR #51, opened 2016-08-18
Provide simple webhook support for statistics

benlovell:
This simply calculates the averages for fulfilment timings and the
window of time between the appointment date and the customer's first
selected slot -- per booking location.

This is then sent POSTed to the configured `STATISTICS_WEB_HOOK_URI` as
a JSON payload. Upon receipt of a non-successful status code it throws
an error. This would be handled during retries, since this would be
scheduled to execute in the context of a sidekiq worker.

Timeouts for the hook can be configured using
`STATISTICS_WEB_HOOK_TIMEOUT` and `STATISTICS_WEB_HOOK_OPEN_TIMEOUT`.

This can be scheduled with `bundle exec rails runner 'StatisticHooksJob.perform_later'`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,9 @@ gem 'sidekiq'
 gem 'sinatra', require: false
 gem 'bugsnag'
 gem 'newrelic_rpm'
+gem 'faraday'
+gem 'faraday-conductivity'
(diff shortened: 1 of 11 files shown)


dwhenry:
:) :shipit:

PR #52, opened 2016-08-18
Web hook improvements

benlovell:
The google sheets API was returning HTML but we were trying to parse this
as JSON. We can rely on a successful status code, the response body is not
important.

Faraday doesn't coerce options from strings. Derp.


benbarnett:
Yes. 

Yes, lovely.

PR #53, opened 2016-08-22
Update `audited` dependency

benlovell:
This is to make way for changes I am making for the activity feed. To
make audit's entries work polymorphically for activities.

The prior version of `audited` was directly mixed in to
`ActiveRecord::Base` as a module so wasn't (at least easily, without
hacks) possible to latently hook the `after_create` for audit entries.

I'll need to reopen the `Audited::Audit` class to add an `after_create`
handler responsible for the creation of the activity entries.

This may seem a fair bit of juggling just to get audits to work as
activities polymorphically, but since audits will eventually make such
a small part of activity type events and we have auditing working
nicely it seems legit to me.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 ruby IO.read('.ruby-version').strip
 
-gem 'audited-activerecord'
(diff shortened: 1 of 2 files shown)


benbarnett:
👍

PR #54, opened 2016-08-22
Ensure statistics fields are not audited

benlovell:
We don't need to track the changes to these fields and their presence
in the audit history will bust up presentation of the activity feed so
I've excluded them.

I'll remove any existing ones separately as this only affects the
production environment.


benbarnett:
Ah yes 👍

PR #55, opened 2016-08-23
Setting up Babel/es6/Jasmine

Guntrisoft:
@benbarnett


benbarnett:
It's pretty self explanatory I guess but maybe just flesh this commit message out a bit? Keeps the history in line with @benlovell's lovely commits!

Otherwise 💯

Guntrisoft:
@benbarnett updated the commit to be more descriptive!

benbarnett:
Nice one 👍

lib/tasks/analyse_javascript.rake
```diff
@@ -0,0 +1,5 @@
+desc 'Analyse Javascript with JSHint and Javascript Code Style checker'
+task :analyse_javascript do
+  puts 'Running Javascript code analysis...'
+  system('npm run js') || raise('Javascript analysis failed')
```
benlovell: Do we need to raise here when we could rely on the exit code of the `npm` sub-process?

Guntrisoft: In testing this change locally (removing the `raise`) it seemed to continue to run the Jasmine tests afterwards, so leaving this in for now.

PR #58, opened 2016-08-25
[WIP] - Appointments calendar

Guntrisoft:


app/assets/javascripts/application.js
```diff
@@ -1,13 +1,8 @@
...
-// about supported directives.
-//
+//= require jquery
+//= require moment/moment
+//= require fullcalendar/dist/fullcalendar
+//= require fullcalendar-scheduler/dist/scheduler
+
 //= require_tree .
+
+//= require components/Calendar
```
benbarnett: This should probably sit before the `require_tree` stuff?

app/assets/javascripts/components/calendars/appointments.es6
```diff
@@ -0,0 +1,40 @@
+module.exports = {
```
benbarnett: As we'll have multiple calendars for appointments, can we rename this to something like `guider_schedule` or similar?

app/views/appointments/calendar.html.erb
```diff
@@ -0,0 +1,71 @@
...
+        <label for="location_<%= location[1] %>"><input type="checkbox" id="location_<%= location[1] %>" name="location_<%= location[1] %>" value="1" checked> <%= location[0] %></label>
+      </p>
+    <% end %>
+  </div>
+</div>
+
+<script>
+window.PWPLAN = window.PWPLAN || {};
+
+$(function() {
```
benbarnett: Seeing this stuff now (and how much nicer it is having the js separated for the rest of it), it probably makes sense to convert this to a feed/endpoint of sorts. I guess we can chat through this with @benlovell too

benlovell:
This won't be used now.

PR #59, opened 2016-08-25
Don't clobber the `default` task

benlovell:
It was pointed out that the rubocop rake task was no longer running by
default since the addition of the `default` overrides.

The default task was clobbered when specified after the rubocop
declaration. This change loads rubocop explicitly and adds the task to the
`default` group after.


Guntrisoft:
👍

PR #60, opened 2016-08-25
Ping slack when we get a booking request

benlovell:
For funsies.


benbarnett:
Happy Phil

benlovell:
I definitely Phil good about this PR.

PR #61, opened 2016-08-29
Add message activities

benlovell:
Gives the booking manager the ability to leave messages against a
booking request and / or an appointment.

There is no validation and the booking manager can leave a blank
message if they like.

### DIFF
--- a/app/assets/javascripts/poller.js
+++ b/app/assets/javascripts/poller.js
@@ -40,7 +40,7 @@
       var element = $(text);
 
       if (!document.getElementById(element.attr('id'))) {
-        element.hide().insertAfter(this.$messageForm).fadeIn();
+        element.hide().prependTo(this.$poller).fadeIn();
       }
     },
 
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/poller.js
```diff
@@ -36,7 +37,11 @@
     },
 
     insertActivity: function(text) {
-      $(text).hide().prependTo(this.$poller).fadeIn();
+      var element = $(text);
+
+      if (this.$poller.find('#' + element.attr('id')).length == 0) {
```
benlovell: Happy to take a steer on something more concise here.

benlovell:
This is good for review now. @benbarnett @Guntrisoft

app/assets/javascripts/poller.js
benbarnett: As id's are unique to the whole page you could probably just use `getElementById` and then boolean-ise it. Does this seem readable?

```js
if (!!!document.getElementById(element.attr('id')) {
```

benlovell: Nicely.

Guntrisoft:
Only observation is you can submit a blank message but as discussed we'll see if someone does do that - otherwise looking good to me 👍 ✅

PR #63, opened 2016-09-02
Removes 'no activities' when an activity is added

Guntrisoft:


Guntrisoft:
@benbarnett @benlovell

benbarnett:
Clever ever ver

PR #65, opened 2016-09-05
Attempt to correctly case given names

benlovell:
This isn't flawless, but will catch those cases where customers enter
their names in all lower case characters.


Guntrisoft:
💯

PR #68, opened 2016-09-05
Resolve auto/eager loading issues after Rails 5

benlovell:
This was only occurring in production due to the changes in autoloading
introduced since Rails  5.

Moving our lib stuff into `app/lib` resolves this as it is
automatically autoloaded and added to the eager-load paths.


dwhenry:
@benlovell just wanted to check if you considered doing the below instead?
```
config.eager_load_paths << Rails.root.join('lib')
```

benlovell:
@dwhenry yes I did, Rails globs the `app` dir and automatically auto/eager loads what it finds there so I'd rather rely on that default than add further configuration. 👍

PR #75, opened 2016-09-22
Remove Jasmine

benlovell:
We don't have javascript unit-style specs. This behaviour is well
covered using capybara / poltergeist style feature specs.


benbarnett:
raw

PR #78, opened 2016-09-23
Handle incoming bounce webhooks from MailGun

benlovell:
When MailGun sends an inbound bounce webhook we look up the latest
matching booking request from the original recipient's email and create
an activity containing the actual reasoning for the hard bounce.

Any problems raise an error that bubbles up to bug snag.

Requires the `MAILGUN_API_TOKEN` to be set.

screen shot 2016-09-23 at 19 38 36

### DIFF
--- a/app/controllers/mail_gun/drops_controller.rb
+++ b/app/controllers/mail_gun/drops_controller.rb
@@ -14,6 +14,7 @@ module MailGun
         :event,
         :recipient,
         :description,
+        :message_type,
         :timestamp,
         :token,
         :signature
--- a/app/controllers/mail_gun/drops_controller.rb
+++ b/app/controllers/mail_gun/drops_controller.rb
@@ -1,0 +1,23 @@
+module MailGun
+  class DropsController < ActionController::Base
+    def create
+      form = DropForm.new(drop_params)
+      form.create_activity
+
+      head :created
+    end
+
+    private
+
+    def drop_params
+      params.permit(
+        :event,
+        :recipient,
+        :description,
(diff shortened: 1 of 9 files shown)


app/views/activities/_bounce_activity.html.erb
```diff
@@ -0,0 +1,4 @@
+<li class="activity-feed__item t-activity" id="activity-<%= bounce_activity.id %>">
+  <span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span>
+  <b>Delivery failure</b> <em><%= bounce_activity.message %></em> <em class="badge"><%= time_ago_in_words(bounce_activity.created_at) %> ago</em>
```
benbarnett: Do we get the mailgun headers here? It might be useful to state [which message](https://github.com/guidance-guarantee-programme/planner/blob/master/app/lib/mailgun_headers.rb#L11) it was that bounced?

benlovell: I guess if one fails they'd all fail since the email would have been incorrect? These are hard bounces. That said - they do pass the headers so we could be more descriptive. I'll check that out. 👍

benbarnett: That raises another point actually, there are two statuses - [`bounce` and `drop`](https://documentation.mailgun.com/api-webhooks.html#webhooks). 

Should we track both? A bounce could feasibly happen further down the line if the user's mailbox is full, whereas the failure would apply to all emails as you say.

PR #79, opened 2016-09-24
Booking manager hides unfulfillable booking

benlovell:
Adds a button on the fulfilment view so booking manager can choose to hide /
deactivate booking requests that are considered unfulfillable after a period of
time.

These won't affect the calculation of our KPIs since they're only
calculated on the presence of an associated appointment.

screen shot 2016-09-24 at 13 44 44

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -20,6 +20,7 @@ class User < ActiveRecord::Base
 
   def unfulfilled_booking_requests
     booking_requests
+      .active
       .includes(:appointment)
       .where(appointments: { booking_request_id: nil })
   end
--- a/app/views/appointments/new.html.erb
+++ b/app/views/appointments/new.html.erb
@@ -153,6 +153,12 @@
         <%= f.button class: 'btn btn-primary btn-block t-submit' do %>
           Make appointment and notify<br><%= @appointment_form.name %>
         <% end %>
+        <%= link_to 'Hide this Booking Request',
+          @appointment_form.location_aware_booking_request,
+          method: :delete,
+          class: 'btn btn-danger btn-block t-deactivate',
+          data: { confirm: "Are you sure you want to hide #{@appointment_form.name}'s booking request?" }
+        %>
       <% end %>
     </div>
   </div>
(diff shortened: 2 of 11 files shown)


app/views/appointments/new.html.erb
```diff
@@ -153,6 +153,12 @@
         <%= f.button class: 'btn btn-primary btn-block t-submit' do %>
           Make appointment and notify<br><%= @appointment_form.name %>
         <% end %>
+        <%= link_to 'Hide this Booking Request',
+          @appointment_form.location_aware_booking_request,
+          method: :delete,
+          class: 'btn btn-danger btn-block t-deactivate',
+          data: { confirm: 'Are you sure?' }
```
benbarnett: Shall we make this a bit more contextual? 

`Are you sure you want to hide #{@appointment_form.name}'s booking request?`

Or something?

app/models/user.rb
```diff
@@ -20,6 +20,7 @@ class User < ActiveRecord::Base
 
   def unfulfilled_booking_requests
     booking_requests
+      .active
       .includes(:appointment)
       .where(appointments: { booking_request_id: nil })
```
AndrewVos: This feels like business logic that could be extracted

AndrewVos: #without_appointments or something?

benlovell: It's already within `unfulfilled_booking_requests` which to my mind is perfectly descriptive. Does the addition of the `active` scope change this significantly?

app/views/appointments/new.html.erb
benlovell: This makes sense. Will do.

app/models/user.rb
AndrewVos: Wow nope I totally just didn't realise the method name! lgtm

app/views/appointments/new.html.erb
benlovell: screen shot 2016-09-26 at 11 07 47

benlovell: @benbarnett Done.

PR #84, opened 2016-10-14
Allow administrators to switch locations

benlovell:
This allows users with the necessary privileges to switch their booking
location 'ownership' thus allowing them to see bookings / appointments
from any location.

Prior to this the user would have to go to signon and change their
organisation then log out / in.

Note: The chosen organisation will be reverted upon the next sign in,
depending on what organisation is assigned in the signon application.

As you'll see, the styling could do with some loving:

screen shot 2016-10-14 at 11 41 01

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -58,7 +58,7 @@ GEM
     babel-transpiler (0.7.0)
       babel-source (>= 4.0, < 6)
       execjs (~> 2.0)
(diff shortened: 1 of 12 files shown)


Guntrisoft:
As agreed away from here, I will look at the styling/positioning of the dropdown later - apart from that, looks good to me 👍

PR #86, opened 2016-11-04
Permit the `date_of_birth` via the booking API

benlovell:
Clients can now pass an optional date of birth to the booking request
API. This is displayed on the booking fulfilment page.


Guntrisoft:
👍

PR #90, opened 2016-11-18
Ensure emails are not sent incorrectly

benlovell:
If the manager submitted an edit to the appointment without actual
changes the customer would incorrectly receive an email notification.

This change guards against an empty changeset when asking if it should
notify the customer.


benbarnett:
Nice one 👍

PR #96, opened 2016-12-28
Use your local system `phantomjs` binary

benlovell:
As per TAP, this speeds up the test runs locally and avoids problems
with the cross-project, shared gem sets.


Guntrisoft:
Retrospective 👍

PR #99, opened 2017-01-03
Toggle display of hidden booking requests

benlovell:
Booking managers can view booking requests that have been hidden, or
just the regular booking requests.

screen shot 2017-01-03 at 15 13 45


benlovell:
Lovely. :100:

PR #100, opened 2017-01-04
Allow managers to reactivate hidden bookings

benlovell:
Since we now display hidden bookings we should also allow the
booking managers to reactivate previously hidden bookings when necessary.

Also we now log an activity when a booking request's activation changes.

## Reactivate
screen shot 2017-01-04 at 10 39 45

## Activities
screen shot 2017-01-04 at 11 23 09


AndrewVos:
Nice, looks good to me.

PR #108, opened 2017-02-02
Allow filtering by reference or customer

benlovell:
Allows booking managers to search appointments by either the booking
reference or a fuzzy match on the customer's first / last name.

screen shot 2017-02-03 at 08 50 19


benlovell:
@benbarnett Yep, absolutely. 👍

PR #113, opened 2017-02-07
Allow filtering by status

benlovell:
Gives booking managers the ability to search and filter by appointment
status as well as existing criteria.

screen shot 2017-02-08 at 10 45 23


Guntrisoft:
👍

PR #116, opened 2017-02-09
Alter button colour to avoid having multiple primary buttons

Guntrisoft:
screen shot 2017-02-09 at 16 37 11
screen shot 2017-02-09 at 16 37 07

Ideally we want just one primary button per page, so here we
use 'info' buttons (from the bootstrap choice of button types)


Guntrisoft:
:flushed:

PR #118, opened 2017-02-10
Add mailer previews

benlovell:
This lets us easily verify the content and layout of our transactional emails
at various points in the service.


Guntrisoft:
👍

PR #119, opened 2017-02-13
Scheduled customer appointment reminders

benlovell:
We will now send scheduled appointment reminders at 3PM every day. This
will send an email reminder to the customer, containing their
appointment details. We are hoping this causes an uplift in appointment
attendance and in turn reduces the no-show rate.

The system guards against multiple reminder deliveries based on the
presence of the newly introduced `ReminderActivity` which also shows in
the activity feed as a record of this reminder being sent.

For this to run we will need to schedule creation of the
`AppointmentRemindersJob`, that in turn fans out a job per appointment
that needs a reminder sent.

## Email copy ##

screen shot 2017-02-17 at 12 41 23

## Activity Feed ##

screen shot 2017-02-13 at 15 03 10

### DIFF
--- a/app/jobs/appointment_reminder_job.rb
+++ b/app/jobs/appointment_reminder_job.rb
@@ -1,0 +1,11 @@
+class AppointmentReminderJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(appointment)
+    booking_location = BookingLocations.find(appointment.location_id)
+
+    ReminderActivity.from(appointment)
+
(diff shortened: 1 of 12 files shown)


Guntrisoft:
👍

Guntrisoft:
[As a customer I receive a reminder email prior to the appointment](https://trello.com/c/0KptAhZD/3-as-a-customer-i-receive-a-reminder-email-prior-to-the-appointment)

PR #120, opened 2017-02-14
Use the latest phantomjs where possible

benlovell:
This gem loads the phantomjs binary from its own local bin path. This
ensures we get the latest / greatest phantomjs in all environments.

We removed this a while back to instead fallback on the locally sourced
binary since it was causing issues on my local machine, I've since
narrowed this down to `exec` spawning shells 'slowly' and patched the
issue.


Guntrisoft:
👍 💯

PR #121, opened 2017-02-15
[WIP] Replace Morden with Mitcham

benlovell:
Morden is closing temporarily so we must reallocate Morden's child
bookings over to Mitcham.

Depends on: guidance-guarantee-programme/location-register#84


benlovell:
Closing in favour of a simpler solution in guidance-guarantee-programme/location-register#85.

PR #122, opened 2017-02-19
Only notify active booking managers

benlovell:
We were attempting to notify `disabled` (this is GDS-SSO terminology
sadly) booking managers that have had their accounts revoked on our
signon instance. One concrete example being Matt's email account that
has been causing hard bounces since he left the Pension Wise
organisation.


Guntrisoft:
:+1:

PR #124, opened 2017-02-24
Turn off polling under certain conditions

Guntrisoft:
Using the Page Visibility API to determine whether the user
is currently looking (active tab) at the app. If they're not
then we stop polling - making the app the active tab again
will resume polling.

After an hour of polling we turn it off to avoid a user
from polling unnecessarily (say overnight), and keeping
their Signon session alive - polling will resume again if the
user moves to another tab and back again.


Guntrisoft:
[Don't continually poll for activities](https://trello.com/c/LJT7sn1k/24-don-t-continually-poll-for-activities)

app/assets/javascripts/poller.js
```diff
@@ -5,15 +5,45 @@
...
+    },
+
+    handleVisibilityChange: function(e) {
+      return document[this.hidden] ? this.stop() : this.start();
+    },
+
     start: function() {
-      setInterval($.proxy(this.poll, this), this.interval());
+      this.timer = setInterval($.proxy(this.poll, this), this.interval());
+      this.timeOutTimer = setTimeout($.proxy(this.stop, this), this.pollingTimeout);
```
benlovell: `timeoutTimer` would make more sense as timeout isn't two words.

app/assets/javascripts/poller.js
```diff
@@ -5,15 +5,45 @@
     init: function() {
       this.$poller = $('.js-poller');
       this.$messageForm = $('.js-message-form');
+      this.pollingTimeout = 1000 * 60 * 60; // 1 hour
 
       if (this.$poller.length && this.$messageForm.length) {
         this.setLast(this.timestamp());
+        this.visibilitySetup();
         this.start();
+        this.bindEvents();
```
benbarnett: Should `this.bindEvents()` happen before `start()`? Just reads a little odd.

app/assets/javascripts/poller.js
```diff
@@ -5,15 +5,45 @@
...
       }
     },
 
+    visibilitySetup: function() {
+      if (typeof document.hidden !== "undefined") { // Opera 12.10 and Firefox 18 and later support
```
benbarnett: We should probably extract this to a meaningful method name e.g. `if (this.supportsVisibilityAPI())` or something?

app/assets/javascripts/poller.js
```diff
@@ -5,15 +5,45 @@
...
+        this.hidden = "hidden";
```
benbarnett: Can we use single quotes here per the standard? And everywhere else

app/assets/javascripts/poller.js
```diff
@@ -5,15 +5,45 @@
...
+
     start: function() {
-      setInterval($.proxy(this.poll, this), this.interval());
+      this.timer = setInterval($.proxy(this.poll, this), this.interval());
+      this.timeOutTimer = setTimeout($.proxy(this.stop, this), this.pollingTimeout);
+    },
+
+    stop: function() {
+      clearInterval(this.timer);
+      clearInterval(this.timeOutTimer);
```
benbarnett: This should be `clearTimeout()` no?

app/assets/javascripts/poller.js
```diff
@@ -5,15 +5,45 @@
...
 
       if (this.$poller.length && this.$messageForm.length) {
         this.setLast(this.timestamp());
+        this.visibilitySetup();
         this.start();
+        this.bindEvents();
       }
     },
 
+    visibilitySetup: function() {
```
benbarnett: What happens here if none of the browsers are supported?

app/assets/javascripts/poller.js
Guntrisoft: Fixed

app/assets/javascripts/poller.js
Guntrisoft: Yup fair enough - fixed

app/assets/javascripts/poller.js
Guntrisoft: Have changed this around a bit, to handle unsupported browsers of the PageVisibility API

app/assets/javascripts/poller.js
Guntrisoft: Copy and paste fail - no double quotes now.

app/assets/javascripts/poller.js
Guntrisoft: Yup - fixed.

app/assets/javascripts/poller.js
Guntrisoft: As above.

app/assets/javascripts/poller.js
```diff
@@ -5,15 +5,60 @@
...
+
+      return null;
+    },
+
+    bindEvents: function() {
+      var prop = this.getHiddenProp();
+
+      if (prop) {
+        $(document).on(
+          prop.replace(/[H|h]idden/, '') + 'visibilitychange',
```
Guntrisoft: Not 100% happy with this - looks a bit ugly - from here: https://www.html5rocks.com/en/tutorials/pagevisibility/intro/

Maybe something cleaner?

app/assets/javascripts/poller.js
```diff
@@ -33,13 +33,17 @@
...
-          prop.replace(/[H|h]idden/, '') + 'visibilitychange',
+          this.visibilityChangeEvent(prop),
           this.handleVisibilityChange.bind(this)
         );
       }
     },
 
-    isHidden: function() {
+    visibilityChangeEvent: function(prop) {
+      return prop.replace(/[H|h]idden/, '') + 'visibilitychange';
```
benlovell: JS supports flags for regex so you could express this case-insensitively as `/hidden/i`.

Guntrisoft:
I'm half wondering about reducing the support for this feature, to simplify the code.

I believe a certain release of IE10 had the ms prefix, and Chrome 32 had the webkit prefix, and Firefox 17 had the moz prefix...

Guntrisoft:
http://caniuse.com/#feat=pagevisibility

Guntrisoft:
@benbarnett @benlovell Taken the decision to simply the code to just supported the W3C standard for page visibility (and have a hour timeout too).

The amount of browsers who don't support the w3c standard for page visibility API is so minor I don't think it's worth worrying about.

I think both the page visibility and hour timeout will both help equally to reduce unnecessary polling for the app.

app/assets/javascripts/poller.js
```diff
@@ -5,15 +5,37 @@
...
 
       if (this.$poller.length && this.$messageForm.length) {
         this.setLast(this.timestamp());
+        this.bindEvents();
         this.start();
       }
     },
 
+    pageVisibilitySupported: function() {
+      return 'hidden' in document ? true : false;
```
benbarnett: Don't think you need the ternary here? Would `return 'hidden' in document` work?

app/assets/javascripts/poller.js
```diff
@@ -5,15 +5,37 @@
...
+    },
+
+    bindEvents: function() {
+      if (this.pageVisibilitySupported()) {
+        $(document).on('visibilitychange', this.handleVisibilityChange.bind(this));
+      }
+    },
+
+    handleVisibilityChange: function() {
+      return document['hidden'] ? this.stop() : this.start();
```
benbarnett: This feels like a duplicate of line 18 – should we be using the same method to check for support?

app/assets/javascripts/poller.js
Guntrisoft: yup true, fixed.

app/assets/javascripts/poller.js
Guntrisoft: Slightly different things - one testing support, and this one testing the value

Guntrisoft: Moved out into a method

benbarnett:
Cool. I think (and could probably do this in another PR), the module as a whole could do with a bit of a cleanup though – its readability could probably be improved. Can go through this in person.

👍

PR #125, opened 2017-03-02
Allow modifications to booking when fulfilling

benlovell:
We now allow the booking managers to modify the booking request in
place before transitioning to the appointment. The booking request
itself is still considered immutable, so this required a 1:1 mirroring
of booking request attributes to appointment attributes.

This also required some duplication of validation rules between the
resultant appointment and the form object. Once we add the further,
editable attributes on the appointment edit screen I can consolidate
these rules into once place.

screen shot 2017-03-02 at 10 00 54

### DIFF
--- a/app/views/appointments/new.html.erb
+++ b/app/views/appointments/new.html.erb
@@ -2,9 +2,8 @@
   <ol class="breadcrumb">
     <li><a href="<%= root_path %>">Appointment planner</a></li>
     <li><a href="<%= root_path %>">Booking requests</a></li>
-    <li class="active t-name"><%= @appointment_form.name %></li>
+    <li class="active"><%= @appointment_form.name %></li>
   </ol>
-
   <h1>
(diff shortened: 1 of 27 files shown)


benlovell:
[As a booking manager I am able to edit a customers details at the booking request stage so that I have the correct details for the appointment.](https://trello.com/c/0SyxQlgN/13-as-a-booking-manager-i-am-able-to-edit-a-customers-details-at-the-booking-request-stage-so-that-i-have-the-correct-details-for-t)

app/views/appointments/new.html.erb
```diff
@@ -97,11 +61,23 @@
         </div>
       </div>
     </div>
+
+    <h3 class="h4">Requested location:</h3>
+    <p class="lead">
+    <strong>
```
benbarnett: Indentation looks slightly out here

Guntrisoft:
👍

PR #126, opened 2017-03-02
Don't audit when copying attributes

benlovell:
This makes no sense to log and shouldn't be pointed out to the customer
by way of an appointment or booking request activity entry.

Amendments to migrations are usually a bad idea but we're fine here
since this has yet to run and in those environments where it has it can
be deemed idempotent.


benlovell:
[As a booking manager I am able to edit a customers details at the booking request stage so that I have the correct details for the appointment.](https://trello.com/c/0SyxQlgN/13-as-a-booking-manager-i-am-able-to-edit-a-customers-details-at-the-booking-request-stage-so-that-i-have-the-correct-details-for-t)

Guntrisoft:
👍

PR #129, opened 2017-03-02
Use redis for a centralised API cache

benlovell:
Cache locations using the centralised (shared) redis cache in the
staging and production environments. Keys are still expired respective
of their configured TTL.

Before now we relied on the memory store for cache backing, this
doesn't survive restarts so deploys were causing multiple cache misses
until the caches were warmed through.


benlovell:
[Locations cache becomes corrupted in background workers](https://trello.com/c/MP00fdh3/38-locations-cache-becomes-corrupted-in-background-workers)

Guntrisoft:
👍

PR #133, opened 2017-03-03
Use curly apostrophe instead of straight one on Don't know

Guntrisoft:
Pension Wise standard is to use smart quotes/apostrophes where
possible

screen shot 2017-03-03 at 12 26 28


benlovell:
I wasn't angry, I was more disappointed.

PR #134, opened 2017-03-08
Ensure cached entries are evicted

benlovell:
The booking locations API gem now ensures the cache TTL is respected.
Since we had to bump several major versions there were a few changes we
had to accomodate in our tests, due mostly to changes in the stubbed
response JSON provided by the gem for testing purposes.


benlovell:
[Ensure TTL is respected on cached booking location responses](https://trello.com/c/ApuBJ5Ak/47-ensure-ttl-is-respected-on-cached-booking-location-responses)

PR #135, opened 2017-03-08
Bump gems with simple `bundle update`

benlovell:
Mainly for the Rails update that includes a bugfix for db runtime
logging at the `info` loglevel and below, but also bumps any other
unpinned gems that had updates recently.


benlovell:
Deployed and tested in staging to ensure we don't face the same issues with `uglifier`. All good 👍

PR #137, opened 2017-03-10
A new activity to record when an appointment mail gets sent to the customer

jongilbraith:
screen shot 2017-03-10 at 14 21 01

### DIFF
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -59,7 +59,7 @@ class AppointmentsController < ApplicationController
   def notify_customer(appointment)
     return unless appointment.notify?
 
-    Appointments.customer(appointment, booking_location.object).deliver_later
+    AppointmentChangeNotificationJob.perform_later(appointment)
   end
 
   def location_aware_appointment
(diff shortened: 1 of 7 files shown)


jongilbraith:
[As a booking manager I want know if an email has been sent to the customer when their details or appointment details have been updated](https://trello.com/c/28eUiBT3/29-as-a-booking-manager-i-want-know-if-an-email-has-been-sent-to-the-customer-when-their-details-or-appointment-details-have-been-u)

app/controllers/appointments_controller.rb
```diff
@@ -59,7 +59,7 @@ def search_params # rubocop:disable Metrics/MethodLength
   def notify_customer(appointment)
     return unless appointment.notify?
 
-    Appointments.customer(appointment, booking_location.object).deliver_later
+    AppointmentChangeNotificationJob.perform_later(appointment)
```
benlovell: A lot of things happen here. Can we test higher up the overall effects of this call? Take a look at the feature specs and do a quick assertion there that this job is enqueued or preferably that a mail is sent and the associated activity is created.

jongilbraith: Cool, will do. I believe the features are already testing that the mail's sent, so should just need to add something for the activity.

PR #138, opened 2017-03-10
Bump Ruby to 2.4

benlovell:
For the latest GC and general improvements.

2.4 gives us the `Regex#match?` method that can be used in favour of
threequals when wanting a boolean match on a regex and rubocop picked
that up.


jongilbraith:
👍

PR #141, opened 2017-03-16
Expose API to reassign child locations

benlovell:
We need to expose an API that is called in response to a child
location being reassigned to a new booking location in the locations
registry. Since we maintain referential integrity with both the booking
location and the child location here we need to be sure to keep them in
sync with the canonical source of locations.

Once the change is made here we simply flush our whole cache rather
than trying to determine which exact locations need flushing. This is
an API call that will happen very infrequently and only while the
location mergers and closures are happening.

Clients can call the API as below:

`PATCH /api/v1/booking_requests/batch_reassign` with the following
payload:

```javascript
  {
    "booking_location_id": "...", // new booking location
    "location_id": "..." // the location being reassigned
  }
```

### DIFF
--- a/app/controllers/api/v1/booking_requests_controller.rb
+++ b/app/controllers/api/v1/booking_requests_controller.rb
@@ -16,6 +16,15 @@ module Api
         end
       end
 
+      def batch_reassign
+        BatchBookingRequestReassignment.new(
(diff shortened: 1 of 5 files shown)


benlovell:

PR #142, opened 2017-03-20
Clear only cached locations

benlovell:
Due to recent changes to `booking_locations` we now prefix booking
locations when read-through cached by the gem, so in turn we can be more
specific about which cache entries we clear.


benlovell:

PR #144, opened 2017-03-22
Create activities to audit reassignments

benlovell:
It's helpful for booking managers to see when a booking was reassigned
to a new booking location.

screen shot 2017-03-22 at 17 51 32


benlovell:

PR #146, opened 2017-03-24
Reassign requests by booking location

benlovell:
When a booking location is assigned to another, the logic around
reassignment is slightly different. First we need to check if the given
location ID refers to a booking location - if it does we can assume the
booking location is being reassigned - so we only update bookings matching
their booking location ID with the given location ID.

Otherwise we fallback to the previous logic and update the booking location
ID of bookings matching the given location ID. This can be confusing since
bookings can take place at their booking location - in which case the
booking location ID and location ID are the same.

### DIFF
--- a/app/lib/batch_booking_request_reassignment.rb
+++ b/app/lib/batch_booking_request_reassignment.rb
@@ -22,6 +22,14 @@ class BatchBookingRequestReassignment
   end
 
   def affected_bookings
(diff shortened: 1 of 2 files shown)


benlovell:
[As an admin I am able to move a parent location under another parent making it a child location](https://trello.com/c/etXGgBxT/44-as-an-admin-i-am-able-to-move-a-parent-location-under-another-parent-making-it-a-child-location)

PR #147, opened 2017-03-24
Guard against missing locations

benlovell:
When the location isn't bound we blow up in interesting ways and it is
difficult to figure out which location is causing the problem. This
change guards against that and in the case of missing locations
helpfully outputs the location ID for instrumentation.


benlovell:
[Guard against 'missing' locations](https://trello.com/c/2s9m3G0x/4-guard-against-missing-locations)

PR #151, opened 2017-04-06
Permit appointments on the same day

benlovell:
As requested by the CABs. The occassionally have capacity to offer
same-day appointments but we were validating against this scenario.
This change permits same-day appointments to be made from existing
bookings.


benlovell:
[Permit appointments today](https://trello.com/c/mjyCJrbF/108-permit-appointments-today)

PR #152, opened 2017-04-18
Specify reply-to path where available

benlovell:
Booking locations can specify a reply-to path to ensure any booking /
appointment related correspondence is directed to the correct group
alias. The alias will then direct these mailings to the booking
managers responsible for the given location. The mailbox alias and
group membership is to be managed by CITA.


benlovell:
[As a booking manager of a particular location I want to receive the customers email when they reply back to any of their appointment emails](https://trello.com/c/2aOnAcVC/95-as-a-booking-manager-of-a-particular-location-i-want-to-receive-the-customers-email-when-they-reply-back-to-any-of-their-appoint)

PR #155, opened 2017-05-08
Bump sinatra to latest released version

benlovell:
This is used by the sidekiq web control panel and was previously pinned
at a beta release.


Guntrisoft:
👍

PR #157, opened 2017-05-15
Introduce states in place of hidden flag

jongilbraith:


app/controllers/booking_requests_controller.rb
```diff
@@ -1,32 +1,46 @@
 class BookingRequestsController < ApplicationController
   def index
+    @search = BookingRequestsSearchForm.new(search_params)
+    @search.status ||= BookingRequest.statuses.first
```
benlovell: Can we default this in the form class instead to tidy this up please.

app/views/appointments/new.html.erb
```diff
@@ -1,181 +1,218 @@
...
+      <p class="lead">
+        <strong>
+          <a href="https://www.pensionwise.gov.uk/locations/<%= @appointment_form.location_aware_booking_request.location_id %>" class="t-location-name">
+            <%= guard_missing_location(@appointment_form.location_aware_booking_request, :location_name) %>
+          </a>
+        </strong>
+      </p>
+      <hr>
+      <button type="button" class="btn btn-default btn-block booking-request-status-btn js-booking-request-change-state-button t-booking-request-change-state-button">
+        Change booking request state
```
benlovell: Can we refer to it as 'status'.

app/views/booking_requests/index.html.erb
```diff
@@ -28,6 +54,9 @@
             <th>Location</th>
             <th>Slots</th>
             <th>Reference</th>
+            <% if @search.status.blank? %>
```
benlovell: For simplicity sake maybe we should just always display the status regardless.

db/migrate/20170518162225_add_status_to_booking_requests.rb
```diff
@@ -0,0 +1,10 @@
+class AddStatusToBookingRequests < ActiveRecord::Migration[5.0]
+  def change
+    add_column :booking_requests, :status, :integer, null: false, default: 0
+
+    BookingRequest.where(active: true).update_all(status: 'active')
+    BookingRequest.where(active: false).update_all(status: 'hidden')
+
+    remove_column :booking_requests, :active, :boolean, null: false, default: true
```
benlovell: No need for the other options you've passed here. They're ignored anyway.

spec/features/booking_manager_views_booking_requests_spec.rb
```diff
@@ -44,10 +60,15 @@ def then_they_see_the_administrative_location_choices
   end
 
   def and_there_are_booking_requests_for_their_location
-    create_list(:hackney_booking_request, 11)
+    @active_booking_requests = create_list(:hackney_booking_request, 8)
+    @active_booking_requests << create(:hackney_booking_request, name: 'Morty Smith')
+    @active_booking_requests << create(:hackney_booking_request, name: 'Jerry Smith')
+    @active_booking_requests << create(:booking_request, name: 'Dalston Dave',
```
benlovell: Can we put each argument on a new line and indented two characters? With the closing paren inline with the @ above.

app/controllers/booking_requests_controller.rb
jongilbraith: I to and fro'd on this and settled on here because ultimately it's for the purpose of the interface that it needs to default to active and not blank/all.

Happy to move it into the form object though as agreed, it's neater there.

app/views/appointments/new.html.erb
jongilbraith: Ah yes, good call, probably the result of some copypasta from the appointments form.

app/views/booking_requests/index.html.erb
jongilbraith: Yep, can do.

db/migrate/20170518162225_add_status_to_booking_requests.rb
jongilbraith: Sure, figured it was good practice to include them for rollbackability, even though we're not likely to do it (and we'd have lost all the state data).

spec/features/booking_manager_views_booking_requests_spec.rb
jongilbraith: 👍

jongilbraith:
@benlovell still want to have a final look over this for anything I've missed but your comments have been addressed.

@Guntrisoft also I think still needs a look from you from an accessibility point of view.

Guntrisoft:
Some accessibility feedback:

1. <nav> surrounding the search form is incorrect and should be just made a <div> (same goes for appointments page)
2. Ensure the fields of the search form are long enough to see the longest status (awaiting customer feedback).
3. Results table should have a <caption> tag describing what the table contains inside the <table> and before the <thead> (if you are hiding the caption from sight users then use a <span> inside it with a ‘sr-only’ class as hiding a caption tag itself off screen can have some funky effects in voiceover.
4. <div class="modal-dialog"> should be <div class="modal-dialog" role=“document”>
5. The H4 within the modal should have an ID matching the aria-labelledby in <div class=“modal”> (id=“status-message”)
6. Remove the aria-hidden=“true” on the <div class=“modal”> otherwise screen readers can’t get to it
7. Is it possible for the status update to be made via AJAX so you don’t go back to the booking_request index page after saving a status? i.e. the status dialog closes and you can resume looking at that booking request.

jongilbraith:
Over to you, @Guntrisoft.

1, 2, 3 - left for you
4, 5, 6 - done
7 - as discussed, just updated the redirect to bring you back to the new appointment page

Guntrisoft:
[As a booking manager I can set the status of a booking request](https://trello.com/c/8OwaC4yp/162-as-a-booking-manager-i-can-set-the-status-of-a-booking-request)

PR #162, opened 2017-05-23
Correct casing for consistency

benlovell:
Relatively minor but we labelled 'All statuses' in title case for appointments
so we ought to do the same for bookings.


benlovell:
[As a booking manager I can set the status of a booking request](https://trello.com/c/8OwaC4yp/162-as-a-booking-manager-i-can-set-the-status-of-a-booking-request)

PR #165, opened 2017-05-27
Customised availability per location

benlovell:
Grants booking managers the ability to manage their location's availability by
way of templated schedules. Locations without schedules will return the same,
default availability currently returned by the booking locations API from the
location registry.

A single schedule runs at any given time, and serves up slots based on the
grace period to the end of the current booking window (6 weeks) and upon
creation of a schedule, bookable slots are generated and persisted to be served
via the publicly available bookable slots API, exposed in this application.
When an existing schedule is in play and another is created that would overlap,
the prior schedule's slots are destroyed and replaced by those generated by the
new schedule.

Some things to note/question:

1. 'Modify' is probably the wrong language for schedules now
2. We need a solution for 'regenerating' slots when they are approaching the
end of the generation window (currently hard-coded at 4 months from the date of
generation)
3. The slots are cacheable and we should investigate this
4. The `booking_locations` gem needs rework once this is deployed - in order to
correctly derive slots - now this application serves those slots via its API

Review commit-by-commit. I'll squash this down after review.

### DIFF
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -1,4 +1,10 @@
 class Schedule < ActiveRecord::Base
+  has_many :bookable_slots
+
+  def generate_bookable_slots!
+    BookableSlotGenerator.new(self).call
+  end
+
   def self.current(location_id)
     where(location_id: location_id)
       .order(created_at: :desc)
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -1,0 +1,14 @@
+class Schedule < ActiveRecord::Base
+  def self.current(location_id)
+    where(location_id: location_id)
+      .order(created_at: :desc)
+      .first_or_initialize(location_id: location_id)
+  end
+
+  def self.slot_attributes
+    columns_hash
+      .keys
+      .select { |k| /_[a|p]m\z/ === k } # rubocop:disable Style/CaseEquality
+      .map(&:to_sym)
+  end
+end
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -1,8 +1,14 @@
 class Schedule < ActiveRecord::Base
-  has_many :bookable_slots
+  has_many :bookable_slots, -> { order(:date) }
+
+  alias default? new_record?
 
   def generate_bookable_slots!
     BookableSlotGenerator.new(self).call
+  end
+
+  def bookable_slots_in_window
+    bookable_slots.windowed
   end
 
   def self.current(location_id)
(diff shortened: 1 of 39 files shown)


benlovell:
[As a booking manager I am able to create a schedule for a location so I can manage availability](https://trello.com/c/NZ3b0eOU/158-as-a-booking-manager-i-am-able-to-create-a-schedule-for-a-location-so-i-can-manage-availability)

app/models/schedule.rb
```diff
@@ -0,0 +1,26 @@
...
+    bookable_slots.windowed
+  end
+
+  def self.current(location_id)
+    where(location_id: location_id)
+      .order(created_at: :desc)
+      .first_or_initialize(location_id: location_id)
+  end
+
+  def self.slot_attributes
```
benlovell: Not sure about this, maybe just a heap easier to explicitly define these attributes as a constant array of symbols instead. 

The purpose of this is to easily and collectively refer to the availability associated attributes defined by the schedule - for example `monday_am`, `monday_pm`, `tuesday_am`... etc.

We needed to jump through some hoops in the factory to accommodate too, given those attributes are not defined during migration.

db/migrate/20170525135108_create_bookable_slots.rb
```diff
@@ -0,0 +1,12 @@
+class CreateBookableSlots < ActiveRecord::Migration[5.1]
+  def change
+    create_table :bookable_slots do |t|
+      t.belongs_to :schedule, null: false, foreign_key: true, index: true
+      t.date :date, null: false
+      t.string :start, null: false
+      t.string :end, null: false
```
benlovell: The naming of this has caused one or two issues given it's a reserved word in Ruby. I've worked around it by qualifying the references and to maintain consistency with the slots generated by the `DefaultBookableSlots` generator but am thinking of renaming instead...

spec/factories/bookable_slots.rb
```diff
@@ -0,0 +1,16 @@
+FactoryGirl.define do
+  factory :bookable_slot do
+    schedule
+    date { Time.zone.now }
+
+    trait :am do
+      start BookableSlot::AM.start
+      add_attribute(:end, BookableSlot::AM.end)
```
benlovell: This is necessary as a work around for the `end` attribute being a reserved word in Ruby.

spec/factories/schedules.rb
```diff
@@ -0,0 +1,18 @@
...
+  factory :schedule do
+    # Hackney
+    location_id 'ac7112c3-e3cf-45cd-a8ff-9ba827b8e7ef'
+
+    trait :blank do
+      begin
+        Schedule.slot_attributes.each do |attribute|
+          add_attribute(attribute, false)
+        end
+      rescue ActiveRecord::StatementInvalid => e
```
benlovell: This happens during migrations (though totally benign) since `Schedule#slot_attributes` causes a database read and the factories are evaluated early on.

PR #166, opened 2017-05-30
Move the accessibility question closer to the phone number

Guntrisoft:
Feedback from booking managers has indicated that
sometimes customers are not asked about their accessibility
requirements when a booking is made because the manager has
missed that the accessibility question is checked.

This brings the accessibility question closer to the phone
number field, so it's more noticeable.

screen shot 2017-05-30 at 10 04 52

### DIFF
--- a/app/views/appointments/new.html.erb
+++ b/app/views/appointments/new.html.erb
@@ -138,29 +138,16 @@
             <div class="form-group">
               <%= f.label :phone %>
               <%= f.text_field :phone, class: 'form-control t-phone' %>
+
+              <%= f.label :accessibility_requirements, class: 'checkbox-inline checkbox-inline--accessibility' do %>
+                <%= f.check_box :accessibility_requirements, class: 't-accessibility-requirements' %> Accessibility requirements?
+              <% end %>
             </div>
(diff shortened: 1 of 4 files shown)


Guntrisoft:
[As a booking manager I want it to be clearer to me that a customer has accessibility requirements when a booking request comes in](https://trello.com/c/yArstnt2/178-as-a-booking-manager-i-want-it-to-be-clearer-to-me-that-a-customer-has-accessibility-requirements-when-a-booking-request-comes-i)

app/views/appointments/new.html.erb
```diff
@@ -200,6 +187,19 @@
                 ) %>
               </div>
             </div>
+
+            <div class="form-group">
+              <p><b>Defined contribution pot confirmed?</b></p>
+              <%= f.label :defined_contribution_pot_confirmed, value: true, class: 'radio-inline' do %>
+                  <%= f.radio_button :defined_contribution_pot_confirmed, true, class: 't-defined-contribution-pot-confirmed-yes' %>
```
benlovell: These two lines needs are indented two spaces too far.

PR #167, opened 2017-05-30
Fix code indenting of radio button/label

Guntrisoft:


AndrewVos:
:medal_military: :medal_sports:

PR #169, opened 2017-06-02
Add an additional info field to booking request and appointments

Guntrisoft:
This allows the API to accept an additional info field which
is then displayed as readonly on the booking request and
appointment edit forms.

screen shot 2017-06-02 at 16 34 27

### DIFF
--- a/app/controllers/api/v1/booking_requests_controller.rb
+++ b/app/controllers/api/v1/booking_requests_controller.rb
@@ -51,6 +51,7 @@ module Api
           :accessibility_requirements,
           :marketing_opt_in,
           :defined_contribution_pot_confirmed,
+          :additional_info,
           slots: %i(date from to priority)
         ).tap { |p| p[:slots_attributes] = p.delete(:slots) }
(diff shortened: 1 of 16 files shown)


Guntrisoft:
[As a customer booking a face-to-face appointment I am able to provide additional information that I feel may help when booking](https://trello.com/c/335eC4WV/186-as-a-customer-booking-a-face-to-face-appointment-i-am-able-to-provide-additional-information-that-i-feel-may-help-when-booking)

db/migrate/20170602144526_add_additional_info_to_booking_request.rb
```diff
@@ -0,0 +1,5 @@
+class AddAdditionalInfoToBookingRequest < ActiveRecord::Migration[5.1]
+  def change
+    add_column :booking_requests, :additional_info, :string
```
benlovell: These ought to be combined into a single migration. Also they should not be nullable and they should default to an empty string.

PR #173, opened 2017-06-16
Display time zone information in customer emails

jongilbraith:
# Booking request confirmation

screen shot 2017-06-16 at 13 51 32

# Appointment confirmation / updated / reminder

screen shot 2017-06-16 at 13 51 05

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -44,6 +44,10 @@ class Appointment < ActiveRecord::Base
     super || Time.zone.now
   end
 
+  def timezone
+    proceeded_at.in_time_zone('London').utc_offset.zero? ? 'GMT' : 'BST'
+  end
+
   def self.needing_reminder
     pending
       .where(proceeded_at: Time.zone.now..48.hours.from_now)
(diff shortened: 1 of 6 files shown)


app/models/appointment.rb
```diff
@@ -44,6 +44,10 @@ def created_at
     super || Time.zone.now
   end
 
+  def gmt_or_bst
```
benlovell: Maybe call this`timezone` instead? Also it should return the string literal 'GMT' not 'UTC'. Nobody will know what UTC is.

benlovell: Also can we add a test for this method? Soz.

jongilbraith: Doh, not paying attention - had meant to use GMT and not UTC. And I've use UTC instead od BST too.

Re: method name, yeah, I was trying to think of a noun to describe it and couldn't think of anything decent. Even `timezone` feels like it could be misconstrued, but I suppose considered in context it's accurate. Will change.

jongilbraith:
[As a customer who has booked an appointment abroad I am aware that the appointment time is in BST.](https://trello.com/c/qG2IqNKI/197-as-a-customer-who-has-booked-an-appointment-abroad-i-am-aware-that-the-appointment-time-is-in-bst)

PR #175, opened 2017-06-29
Customers may chose between one and three slots

benlovell:
When we have actual availability we shouldn't force customers to chose
three slots. This change permits the customer to select between one and
three potential slots.

booking requests

make appointment


benlovell:
[As a customer when booking an appointment I am able submit the booking form even if I haven't requested three slots](https://trello.com/c/u18jebp4/160-as-a-customer-when-booking-an-appointment-i-am-able-submit-the-booking-form-even-if-i-havent-requested-three-slots)

PR #176, opened 2017-06-29
Include number of chosen slots for slack ping

benlovell:
Pings slack with the location and number of slots the customer chose.


benlovell:
[As a customer when booking an appointment I am able submit the booking form even if I haven't requested three slots](https://trello.com/c/u18jebp4/160-as-a-customer-when-booking-an-appointment-i-am-able-submit-the-booking-form-even-if-i-havent-requested-three-slots)

PR #177, opened 2017-06-30
Don't display schedules for hidden locations

benlovell:
Booking managers seem to be confused about how to treat hidden
locations so some have schedules and some don't. This isn't necessary
since they don't appear to customers during the booking journey and
therefore cannot be booked.

Some locations have a great number of hidden locations and picking
through them when managing schedules is a pain, so we elected to not
display them in the schedules grid.

### DIFF
--- a/app/controllers/schedules_controller.rb
+++ b/app/controllers/schedules_controller.rb
@@ -1,6 +1,7 @@
 class SchedulesController < ApplicationController
   def index
     @booking_location = booking_location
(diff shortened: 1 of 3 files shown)


benlovell:
[As a booking manager creating a schedule for a location I am not able to see the 'hidden' locations in the list](https://trello.com/c/v55vLz4k/214-as-a-booking-manager-creating-a-schedule-for-a-location-i-am-not-able-to-see-the-hidden-locations-in-the-list)

PR #180, opened 2017-07-18
Regenerate scheduled slots approaching end

benlovell:
When scheduled daily, this job will pick up those schedules with
availability ending on the day of execution and regenerate slots from that
day on to 4 months from then based on the slots in the given schedule.

It's not expected that there will be many schedules per day this affects
given the variance in schedules and the day their availability is scheduled
to end.


benlovell:
[Schedules must regenerate slots when approaching the booking window threshold](https://trello.com/c/JcBQvvmy/185-schedules-must-regenerate-slots-when-approaching-the-booking-window-threshold)

PR #182, opened 2017-07-26
Fix date-sensitive spec

benlovell:
Towards the end of a month this spec could travel forward into the next,
hence the slots required would not be visible on the current month.
Travelling to the middle of the month ensures this never happens.


benlovell:
[Fix fragile, date-sensitive specs for appointment counts on the availability calendar](https://trello.com/c/oHrFCfQC/247-fix-fragile-date-sensitive-specs-for-appointment-counts-on-the-availability-calendar)

PR #187, opened 2017-08-22
Add flag to identify bookings placed by agents

benlovell:
This allows us to report on bookings placed by TP agents using the same
process as the regular customer journey.


benlovell:
[As an analyst I am able to identify which face to face appointments are booked online by TP agents](https://trello.com/c/IAD9T41A/262-as-an-analyst-i-am-able-to-identify-which-face-to-face-appointments-are-booked-online-by-tp-agents)

PR #188, opened 2017-09-14
Determine the canonical number for bookings

benlovell:
Allows the administrator to override a parent's online booking twilio
number with that of its child location.


benlovell:
[As a pension wise admin when I assign a child location an 'online booking twillio number' this is displayed in customer emails](https://trello.com/c/1xnIBIwE/273-as-a-pension-wise-admin-when-i-assign-a-child-location-an-online-booking-twillio-number-this-is-displayed-in-customer-emails)

PR #190, opened 2017-09-19
Remove `includes`

benlovell:
This was nerfing the ordering on slots when requested through the
associated booking request.


benlovell:
[Ensure slot selections are presented in order of preference](https://trello.com/c/WI0whueK/281-ensure-slot-selections-are-presented-in-order-of-preference)

PR #191, opened 2017-10-11
Restrict appointment times from 8AM to 6:45PM

benlovell:
This stops booking managers choosing times outside of those servicable
by the various locations.


benlovell:
[As a booking manager fulfilling an booking request I am only able to choose a time between 08:00AM and 18:00PM.](https://trello.com/c/xkKYfoFP/302-as-a-booking-manager-fulfilling-an-booking-request-i-am-only-able-to-choose-a-time-between-0800am-and-1800pm)

PR #194, opened 2017-10-30
Highlight empty schedules

benlovell:
This provides the booking managers a broad view of availability across
their locations.


benlovell:
[As a booking manager I need to know when the availability calendar no longer shows a recurring schedule.](https://trello.com/c/xs6tN5Rk/323-as-a-booking-manager-i-need-to-know-when-the-availability-calendar-no-longer-shows-a-recurring-schedule)

PR #197, opened 2017-11-02
Remove CITA slots over the Christmas period

benlovell:
These have likely been created by schedules and regeneration of
existing schedules that intersect with this period. Since the CITA
locations have blanket closures over this period we should remove all
available slots.


benlovell:
[As a booking manager, I need the Bank Holidays added to Planner as mandatory closure days.](https://trello.com/c/bkizK0Ca/327-as-a-booking-manager-i-need-the-bank-holidays-added-to-planner-as-mandatory-closure-days)

PR #200, opened 2017-11-20
Clarify what happens when saving a schedule

benlovell:
This makes it clearer that a new schedule will overwrite any existing
availability created manually by a booking manager.


benlovell:
[As a Booking Manager I need to confirm the creation of a new schedule](https://trello.com/c/o8oVo6oJ/339-as-a-booking-manager-i-need-to-confirm-the-creation-of-a-new-schedule)

PR #202, opened 2017-11-23
Add rake task for redacting customer details

benlovell:
Adds a rake task that accepts a booking reference and removes any
identifying customer data.


benlovell:
[As a customer, I would like my personal details removed from the Planner system upon request](https://trello.com/c/jZXEwt8o/343-as-a-customer-i-would-like-my-personal-details-removed-from-the-planner-system-upon-request)

PR #205, opened 2017-11-27
Show postal confirmation message at booking stage

benlovell:
This was previously only being displayed at the appointment stage.


benlovell:
[As a TP agent I am able to request an appointment for a customer](https://trello.com/c/WRq8UUhd/318-as-a-tp-agent-i-am-able-to-request-an-appointment-for-a-customer)

PR #206, opened 2017-11-28
Correctly validate email presence

benlovell:
In the case of agent bookings we should not require an email here when
an address is provided during the booking request phase.


benlovell:
[Correct postal/email confirmation validation for agent bookings](https://trello.com/c/LJP1BGdH/346-correct-postal-email-confirmation-validation-for-agent-bookings)

PR #207, opened 2017-11-29
Permit 500 characters for `additional_info`

benlovell:
This wasn't changed in step with its booking request counterpart.


benlovell:
[Extend appointment's additional notes limit](https://trello.com/c/JBVn2YT1/347-extend-appointments-additional-notes-limit)

PR #208, opened 2017-12-01
Disallow `tpbookings@pensionwise.gov.uk`

benlovell:
Some agents are using this as a workaround to avoid entering an address.


benlovell:
[As a Teleperformance agent, I cannot use the tpbookings@pensionwise.gov.uk address when using the new booking form](https://trello.com/c/EEtfRJEh/349-as-a-teleperformance-agent-i-cannot-use-the-tpbookingspensionwisegovuk-address-when-using-the-new-booking-form)

PR #209, opened 2017-12-05
Disable the button during submission

benlovell:
Hopefully this will stop some of the duplicate bookings being made by
agents.


benlovell:
[Disable the submission button while the form is processed](https://trello.com/c/sLeUDdLQ/355-disable-the-submission-button-while-the-form-is-processed)

PR #213, opened 2017-12-20
Alter confirmation message for new year

benlovell:
Ensures the customer is informed their request will not be processed
until the new year.


benlovell:
[As a customer I should be made aware of delays to the booking process over the Christmas period](https://trello.com/c/JFvvkIVS/364-as-a-customer-i-should-be-made-aware-of-delays-to-the-booking-process-over-the-christmas-period)

PR #216, opened 2017-12-27
Provide an agent booking search screen

benlovell:
Gives agents the ability to search for booking requests, only placed by
agents. They will not see normal booking requests from customers.


benlovell:
[As a TelePerformance team leader, I need to know the details of booking requests submitted by TP agents.](https://trello.com/c/oVkML6Nu/366-as-a-teleperformance-team-leader-i-need-to-know-the-details-of-booking-requests-submitted-by-tp-agents)

PR #217, opened 2017-12-29
Revert "Alter confirmation message for new year"

benlovell:
This no longer applies after the new year break.


benlovell:
[As a customer I should be made aware of delays to the booking process over the Christmas period](https://trello.com/c/JFvvkIVS/364-as-a-customer-i-should-be-made-aware-of-delays-to-the-booking-process-over-the-christmas-period)

PR #218, opened 2017-12-29
Notify when status changes to pending

benlovell:
When an appointment changed from cancelled back to pending the customer
was not being informed and thus appointments were being missed.

This change ensures customers are notified by email when the status
changes back to pending from any other.

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -32,7 +32,10 @@ class Appointment < ActiveRecord::Base
   end
 
   def notify?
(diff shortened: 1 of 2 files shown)


benlovell:
[As a booking manager, I require customers to receive appointment confirmation/reminder emails if a cancelled appointment is rescheduled](https://trello.com/c/QCh7qMHX/365-as-a-booking-manager-i-require-customers-to-receive-appointment-confirmation-reminder-emails-if-a-cancelled-appointment-is-resch)

PR #220, opened 2018-01-13
Stop polling upon receipt of an erroneous status

benlovell:
This ensures the poller stops when it receives an error. This has been
happening when booking requests can no longer be found. Due to the
polling interval this will cause a spike of reported errors
unnecessarily.

### DIFF
--- a/app/assets/javascripts/poller.js
+++ b/app/assets/javascripts/poller.js
@@ -43,8 +43,15 @@
     },
 
     poll: function() {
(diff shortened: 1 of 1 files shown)


benlovell:
[Automatically unsubscribe pollers when the activity feed returns an erroneous response](https://trello.com/c/odsbsJDg/375-automatically-unsubscribe-pollers-when-the-activity-feed-returns-an-erroneous-response)

PR #222, opened 2018-01-26
Resolve agent booking search issues

benlovell:
We don't need to pre-select the status field and when no agent was
provided we were still adding an agent clause checking against a blank
ID, thus causing incorrect results.


benlovell:
[Do not pre-select 'status' option when search for agents](https://trello.com/c/tCSNdS2x/390-do-not-pre-select-status-option-when-search-for-agents)

PR #224, opened 2018-02-06
Ensure DOB is validated before eligibility

benlovell:
This was causing errors to be raised at the preview stage and the
resulting message is not particularly useful to the agent.


benlovell:
[Improve validation for DOB field during agent booking preview](https://trello.com/c/twZDcmKv/396-improve-validation-for-dob-field-during-agent-booking-preview)

PR #226, opened 2018-02-14
Exclude guider conference days

benlovell:
Ensure these slots are not returned.


benlovell:
[As a booking manager I need the Pension Wise Summer School added as mandatory closure dates on Planner and TAP](https://trello.com/c/3ThYWz9w/405-as-a-booking-manager-i-need-the-pension-wise-summer-school-added-as-mandatory-closure-dates-on-planner-and-tap)

PR #227, opened 2018-02-20
Allow agents to pick DoB with a control

benlovell:
Gives agents the ability to use a date picker when selecting the
customer's date of birth. This should help avoid the errors in
formatting that are regularly occuring.


benlovell:
[As an agent I want a friendly date picker when providing a customer's date of birth](https://trello.com/c/IVoq2oZR/411-as-an-agent-i-want-a-friendly-date-picker-when-providing-a-customers-date-of-birth)

PR #231, opened 2018-03-27
Update confirmation/reminder emails

benlovell:
Improves this content for both the HTML and plain text versions.


benlovell:
[Amend TAP / F2F email templates so that it refers to 'special feature' and 'state benefits'](https://trello.com/c/m7OvHLwr/457-amend-tap-f2f-email-templates-so-that-it-refers-to-special-feature-and-state-benefits)

PR #232, opened 2018-04-03
Block out guider induction day for CANI

benlovell:
Ensure slots cannot be created on this day.


benlovell:
[Ensure CANI locations are during 9th April](https://trello.com/c/R0xp9t3d/468-ensure-cani-locations-are-during-9th-april)

PR #234, opened 2018-04-20
Batch up data on location availability

benlovell:
Roll this daily snapshot up into something easily queryable from the
blazer reporting platform. This should be scheduled to run at 8AM daily.


benlovell:
[As a research or business analyst, I require access to reporting which indicates the 'availability' of appointments](https://trello.com/c/GnF4FDHV/393-as-a-research-or-business-analyst-i-require-access-to-reporting-which-indicates-the-availability-of-appointments)

PR #236, opened 2018-04-26
Add the ability to edit the postal address for confirmation

pixeltrix:
Agents often make mistakes when entering the postal addresses so add an edit option for both before and after fulfilment so that mistakes can be correct without a developer having to be involved.


app/views/appointments/edit.html.erb
```diff
@@ -79,8 +79,9 @@
       <% if @appointment_form.postal_confirmation? %>
         <hr>
       <h2 class="h3">Postal address</h2>
+      <%= simple_format(@appointment_form.postal_address_lines, class: 't-postal-address') %>
       <p>
-        <%= simple_format(@appointment_form.postal_address_lines) %>
+        <small><%= link_to 'Edit', edit_booking_request_postal_address_path(@appointment_form.booking_request), class: 't-edit-postal-address' %></small>
```
pixeltrix: The `simple_format` method wraps the text with a `<p>` tag by default so this was actually creating invalid HTML before.

config/puma.rb
```diff
@@ -3,7 +3,7 @@
 rackup      DefaultRackup
 daemonize   false
 environment ENV.fetch('RACK_ENV', 'development')
-workers     ENV.fetch('WEB_CONCURRENCY', 2)
+workers     ENV.fetch('WEB_CONCURRENCY', 2) unless Rails.env.development?
```
benlovell: 👍

app/views/postal_addresses/edit.html.erb
```diff
@@ -0,0 +1,69 @@
...
+              <li><%= msg %></li>
+            <% end %>
+          </ul>
+        </div>
+      <% end %>
+      <div class="row">
+        <div class="col-md-12">
+          <div class="form-group">
+            <%= f.label :address, for: :postal_address_address_line_one %>
+            <%= f.text_field :address_line_one, class: 'form-control t-address-line-one' %>
```
benlovell: Can these be placed in separate form groups so they are spaced out a little? I see the intention here but elsewhere in the platform, wherever we take address lines we space the fields consistently.

PR #238, opened 2018-04-27
Don't permit excluded dates for custom slots

benlovell:
This ensures excluded days (holidays and other special internal events) are
not valid days for customised availability.


benlovell:
[Ensure that mandatory closure dates for Pension Wise Summer School cannot be overwritten when setting custom availability](https://trello.com/c/mptk9VU1/501-ensure-that-mandatory-closure-dates-for-pension-wise-summer-school-cannot-be-overwritten-when-setting-custom-availability)

PR #240, opened 2018-04-30
Notify booking manager when email dropped

pixeltrix:


pixeltrix:
@benlovell I was unsure whether you wanted two separate code paths for failures from the original booking request email and failures from confirming the appointment so if you do let me know. Also do we want to try to give some idea about the error? e.g. If their mailbox is full, etc.

pixeltrix:

PR #241, opened 2018-05-01
Display the agent landing page

benlovell:
When the user only possesses the `agent` permission they should be
displayed a notice and link back to the customer booking journey.


benlovell:
[As an agent I need to login to the planner application successfully](https://trello.com/c/jKADfHqi/502-as-an-agent-i-need-to-login-to-the-planner-application-successfully)

PR #242, opened 2018-05-01
Paginate booking location schedules

pixeltrix:
Some booking locations such as Drumchapel have a large number of child locations so paginate the array returned from the JSON request.


pixeltrix:
[Add pagination to Planner Schedules page](https://trello.com/c/gSPvf4r5/464-add-pagination-to-planner-schedules-page)

PR #243, opened 2018-05-01
Lookup booking managers by DC

benlovell:
This was previously looking up the associated booking managers by the
child location's ID, not the delivery centre's ID.


benlovell:

PR #245, opened 2018-05-02
Store status changes separately for reporting

benlovell:
This helps us better understand the various states and what triggers
them during the appointment lifecycle. Previously, this information was
locked away in the audit trail and was not easily queried.

This also includes a rake task to backfill previous appointments with
their associated status transitions.


benlovell:

PR #246, opened 2018-05-02
Send both a 48hr and a 7-day email reminder

benlovell:
Previously we had just sent an appointment reminder email at 48hrs
before the appointment. Sending an earlier reminder at 7-days prior to
the appointment is thought to help mitigate cancellations and no-shows
thus freeing up slots earlier for other customers.

I removed the check for an existing `ReminderActivity` instance as this
just adds further complexity for no apparent gain. The reminders have
been running for over a year and have never caused issues with regards
retries or any circumstances that would cause a customer to receive
multiple, erroneous reminders.


benlovell:

PR #247, opened 2018-05-02
Add push notifications for email failures

pixeltrix:
Using in-page growl type notifications to alert booking managers to email failures. Notifications aren't automatically dismissed but don't persist across page loads.


pixeltrix:

pixeltrix:
Example notification:

pusher

PR #250, opened 2018-05-02
Fix the 2-day range for reminders

benlovell:
Now the job is moving to run once per day we should fix the range to
ensure the correct appointments are included for reminders.


benlovell:

PR #253, opened 2018-05-21
GDPR customer research consent

benlovell:
We are supplanting the marketing opt in with GDPR compliant contact
preferences. We migrate away from the prior marketing opt in to the new
'customer research consent' answer. This is across the API and the
agent-based booking journey.


benlovell:
[Update F2F TP Booking form](https://trello.com/c/zyTLTa58/534-update-f2f-tp-booking-form)

PR #254, opened 2018-05-25
Display customer's selected GDPR consent

benlovell:
CAB's need to see this to know what to enter into their own systems. We now
display the customer's preferences against both the booking request and
appointment.

screen shot 2018-05-25 at 10 16 18


benlovell:
[Display GDPR customer consent](https://trello.com/c/lsfi4bK3/549-display-gdpr-customer-consent)

PR #256, opened 2018-05-30
Allow weekend availability when permitted

benlovell:
When the location's `online_booking_weekends` flag is set, we need to
display Saturday/Sunday so the booking manager can assign availability
for those days. When availability exists across the weekend, this is
automatically presented for booking in the customer and agent journeys.


benlovell:
[Saturday availability](https://trello.com/c/QAbHVXL0/550-saturday-availability)

PR #257, opened 2018-05-30
Correctly validate when no slots are chosen

benlovell:
Age and appointment type require at least one slot to be chosen to
correctly calculate the age at time of appointment. We need to guard
against instances when agents attempt to submit a booking without first
selecting at least one slot.

### DIFF
--- a/app/forms/agent_booking_form.rb
+++ b/app/forms/agent_booking_form.rb
@@ -93,7 +93,7 @@ class AgentBookingForm # rubocop:disable ClassLength
   end
 
   def age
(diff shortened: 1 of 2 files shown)


benlovell:
[As a customer I am able to book an appointment even though I am under 50 but will be eligible by the time I have the appointment.](https://trello.com/c/g8XAEwvC/527-as-a-customer-i-am-able-to-book-an-appointment-even-though-i-am-under-50-but-will-be-eligible-by-the-time-i-have-the-appointment)

PR #258, opened 2018-06-01
SMS reminders and cancellations

benlovell:
Handle SMS cancellation requests

The customer can reply to their SMS reminder and cancel the appointment.
When this happens, the appointment is cancelled, the activity feed is
updated and the booking managers are notified.

Send SMS reminders 48hr before appointment

Sends an SMS reminder to the customer at 48hrs before their appointment is
due to proceed. Adds an activity in the feed for tracking, also allows the
customer to reply 'Cancel' for cancellation.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,7 @@ source 'https://rubygems.org' do # rubocop:disable Metrics/BlockLength
   gem 'kaminari'
   gem 'momentjs-rails'
   gem 'newrelic_rpm'
+  gem 'notifications-ruby-client'
   gem 'oj'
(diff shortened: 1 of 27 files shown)


benlovell:
[As a customer who has booked a F2F appointment I receive a reminder SMS and can cancel the appointment via response](https://trello.com/c/nud35pRj/354-as-a-customer-who-has-booked-a-f2f-appointment-i-receive-a-reminder-sms-and-can-cancel-the-appointment-via-response)

benlovell:
Should be merged with guidance-guarantee-programme/pension_guidance#1082

PR #259, opened 2018-06-06
Display validation error for date of birth

benlovell:
When determing age/eligibility and the agent enters an incorrectly
formatted date of birth eg '01/19/1950' they are shown an error message
and not the more helpful guidance from the validation message. This
change ensures they understand when they've incorrectly formatted a
date of birth.


benlovell:
[Display friendly validation messages for incorrectly formatted date of birth entries](https://trello.com/c/IYInEGaa/564-display-friendly-validation-messages-for-incorrectly-formatted-date-of-birth-entries)

PR #261, opened 2018-06-11
Do not send email reminders upon booking

benlovell:
Guards against the sending of email reminders on the same day as
appointments are booked. We already send a confirmation so the reminder
would be redundant.


benlovell:
[Ensure that telephone appointment & F2F customers do not receive reminder if booking within 2 or 7 day window](https://trello.com/c/4NZA26NG/572-ensure-that-telephone-appointment-f2f-customers-do-not-receive-reminder-if-booking-within-2-or-7-day-window)

PR #262, opened 2018-06-12
Rename from 'Appointment Planner'

benlovell:
We've never referred to this system as 'Appointment Planner' in comms
or otherwise, free up valuable real-estate in the navigation
bar and make the naming consistent with code artifacts.


benlovell:
[Rename 'Appointment Planner'](https://trello.com/c/8L5ebIIV/574-rename-appointment-planner)

PR #263, opened 2018-06-13
Alter GDPR consent when redacting

benlovell:
Also, ensure we correctly filter the attributes when processing any
associated appointment record.


benlovell:
[Process request for customer data redaction](https://trello.com/c/yQ55vVPu/578-process-request-for-customer-data-redaction)

PR #266, opened 2018-06-13
Ensure slots are presented by priority

benlovell:
This has regressed once or twice. Adding this spec here to ensure it
doesn't happen again.


benlovell:
[Ensure that customer appointment choices appear on the booking request screen in order of customer preference](https://trello.com/c/wBLJnTp9/579-ensure-that-customer-appointment-choices-appear-on-the-booking-request-screen-in-order-of-customer-preference)

PR #269, opened 2018-06-28
Do not prefill appointment date during fulfilment

benlovell:
Per the title, booking managers have asked that we don't prefill the
appointment date with the primary slot, during fulfilment.


benlovell:
[Don't preselect appointment date from primary slot during fulfilment](https://trello.com/c/yeD3uRNb/594-dont-preselect-appointment-date-from-primary-slot-during-fulfilment)

PR #271, opened 2018-06-29
Don't preselect slot time either

benlovell:
We recently changed the fulfilment process to not preselect the primary
slot date. Booking managers have further requested we apply the same
logic to the slot time.

We also stop the date field from autocompleting, thus stopping the
prior values from obscuring the JS date picker component when both are
displayed together.


benlovell:
[Don't preselect appointment date from primary slot during fulfilment](https://trello.com/c/yeD3uRNb/594-dont-preselect-appointment-date-from-primary-slot-during-fulfilment)

PR #275, opened 2018-07-05
Display holiday notice to CANI locations

benlovell:
Customers in these locations are being warned that their appointment
may take longer than the usual two days to be scheduled.


benlovell:
[Add public holiday notice for CANI locations](https://trello.com/c/aorMNdlX/604-add-public-holiday-notice-for-cani-locations)

PR #277, opened 2018-07-09
Display booking date/time elsewhere

benlovell:
We display the full booking date/time elsewhere so for consistency and
ease-of-use we should do the same in other booking and appointment views.


benlovell:
[Add booking request timestamp to appointment summary page and TP booking summary pages](https://trello.com/c/uNJdo3J7/609-add-booking-request-timestamp-to-appointment-summary-page-and-tp-booking-summary-pages)

PR #278, opened 2018-07-10
Display the full timestamp of an activity

benlovell:
This is much clearer and precise.


benlovell:
[Add timestamp to customer activity feed](https://trello.com/c/a11j992S/610-add-timestamp-to-customer-activity-feed)

PR #281, opened 2018-07-18
Validate time presence

benlovell:
Since we removed the preselection of time during the fulfilment of
bookings, we've seen several instances where agents or booking managers
forget to select the appointment time. This causes errors further down
in the mapper since their is an attempt to format the time.

Adding this validation will stop that from occuring and inform the
person attempting to place the booking exactly where they have gone
wrong.


benlovell:
[Validate time is supplied during fulfilment](https://trello.com/c/Uo8cqtSo/627-validate-time-is-supplied-during-fulfilment)

PR #282, opened 2018-07-25
Add some buffer to the connection pools

benlovell:
`MAX_THREADS` is set to 10 in production. We often exhaust the pool in
the background workers, this shouldn't happen as sidekiq has a
concurrency of 10. Increasing the pool size should give us some extra
room but I'll keep an eye on it.


benlovell:
[Increase DB pool size for background workers](https://trello.com/c/0A0gYSe7/632-increase-db-pool-size-for-background-workers)

PR #283, opened 2018-07-25
Merge settings for DB configuration

benlovell:
Whoops, we weren't merging the default settings prior to this, so the
default pool sizes were being used.


benlovell:
[Increase DB pool size for background workers](https://trello.com/c/0A0gYSe7/632-increase-db-pool-size-for-background-workers)

PR #284, opened 2018-07-25
Add activity entry when resending emails

benlovell:
This was previously being skipped as the appointment could be
'updated', but this path couldn't have been followed anyway, always
resulting in the 'booked' activity - which is good enough in this
instance.


benlovell:
[Log an activity when resending a confirmation at the appointment stage](https://trello.com/c/T3xuzAQ8/635-log-an-activity-when-resending-a-confirmation-at-the-appointment-stage)

PR #285, opened 2018-07-26
Add phonetic email cue

benlovell:
This is a trial to see if we can improve on the quality of TP supplied
emails and the respective deliverability of confirmation emails.


benlovell:
[Add agent prompt below email field on TP F2F booking form](https://trello.com/c/G40p59Dm/636-add-agent-prompt-below-email-field-on-tp-f2f-booking-form)

PR #286, opened 2018-08-03
Improve accessibility labelling

benlovell:


benlovell:
[As a Booking Manager I would like customers to share accessibility needs when booking an appointment via the Pension Wise website.](https://trello.com/c/lVi4Mfx5/353-as-a-booking-manager-i-would-like-customers-to-share-accessibility-needs-when-booking-an-appointment-via-the-pension-wise-websit)

PR #287, opened 2018-08-20
Validate time when fulfilling

benlovell:
We had a few instances where appointments were created for times outside
of the usual office hours of 8AM to 7PM. This change ensures this does
not happen again.


benlovell:
[Do not allow appointments to be created outside of the permitted hours - 8AM to 7PM](https://trello.com/c/Dzbap0UC/653-do-not-allow-appointments-to-be-created-outside-of-the-permitted-hours-8am-to-7pm)

PR #288, opened 2018-08-23
Reduce grace period across the weekend

benlovell:
Customers will now see slots for Wednesday on Saturday and Sunday,
previously the earliest would have been Thursday.


benlovell:
[Reduce the weekend grace period](https://trello.com/c/fzBR7eof/660-reduce-the-weekend-grace-period)

PR #289, opened 2018-08-28
Send printed confirmations by notify

benlovell:
Sends printed appointment confirmations via GOVUK notify.


benlovell:
[As a face-to-face or telephone appointment customer with no email address I should receive a printed appointment letter via NOTIFY](https://trello.com/c/nesGXxNv/626-as-a-face-to-face-or-telephone-appointment-customer-with-no-email-address-i-should-receive-a-printed-appointment-letter-via-noti)

PR #291, opened 2018-09-20
Favour SMS reminders only where possible

benlovell:
We now match the 7 day reminder behaviour - eg if the customer has a
mobile number we will not send an email reminder, preferring an SMS
reminder. If the customer has no mobile number we will always send email
reminders. So effectively they are now mutually exclusive.

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -78,10 +78,7 @@ class Appointment < ActiveRecord::Base
   end
 
   def self.needing_reminder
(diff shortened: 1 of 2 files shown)


benlovell:
[Replace 2 day reminder email for F2F appointments with sms reminder](https://trello.com/c/AJNbpE4d/672-replace-2-day-reminder-email-for-f2f-appointments-with-sms-reminder)

PR #293, opened 2018-10-09
Only include the guider's first name in letters

pixeltrix:
This matches the email template.


pixeltrix:
[Amend postal appointment/reminder letters so that they only include the guiders first name.](https://trello.com/c/VViTm8Iy/713-amend-postal-appointment-reminder-letters-so-that-they-only-include-the-guiders-first-name)

PR #294, opened 2018-10-15
Do not notify customer when appointment elapsed

benlovell:
We were always sending notifications regardless whether the appointment
had elapsed and this didn't make any sense.

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -54,7 +54,7 @@ class Appointment < ActiveRecord::Base
   end
 
   def notify?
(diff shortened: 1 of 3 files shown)


benlovell:
[Amend configuration of Notify letters so that they are only produced for confirmation and changes to appointments (date/time/venue)](https://trello.com/c/9AVYf2Xv/715-amend-configuration-of-notify-letters-so-that-they-are-only-produced-for-confirmation-and-changes-to-appointments-date-time-venu)

PR #298, opened 2018-11-01
Christmas and other exclusions

benlovell:
Adds exclusions and deletes slots that were created for the Christmas closure
period, except for Drumchapel and its children.


benlovell:
[As a booking manager, I need the 2019 Bank Holidays added to Planner as mandatory closure days.](https://trello.com/c/mPXAjWyr/732-as-a-booking-manager-i-need-the-2019-bank-holidays-added-to-planner-as-mandatory-closure-days)

PR #299, opened 2018-11-08
Skip bank holidays in the grace period

benlovell:
Bank holidays were not accounted for during the grace period previously.
This change ensures we bring parity with TAP (and eventually the other
planners).


benlovell:
[Configure F2F Planner & TAP so the booking window used the same library/logic.](https://trello.com/c/LVpP9vCr/737-configure-f2f-planner-tap-so-the-booking-window-used-the-same-library-logic)

PR #305, opened 2018-11-30
Cancel appointment regardless of validation

benlovell:
We had appointments with overlapping guiders causing the cancellation to
fail as validation would fail with `update!`. We can blindly update the
`status` attribute here regardless of other validations so that's what
we should do.


benlovell:
[Ensure SMS cancellations do not trigger guider validation for overlapping appointments](https://trello.com/c/ZvYpKXTw/754-ensure-sms-cancellations-do-not-trigger-guider-validation-for-overlapping-appointments)

PR #306, opened 2018-11-30
Do not allocate non-realtime bookings

benlovell:
Ensures this validation only runs for realtime bookings during the
crossover period. This can be simplified when all locations are
enrolled.


benlovell:
[Ensure booking request status updates do not needlessly trigger slot allocation validations](https://trello.com/c/mUC1pL1k/756-ensure-booking-request-status-updates-do-not-needlessly-trigger-slot-allocation-validations)

PR #309, opened 2018-12-04
Add a further appointment status

benlovell:
We had a request from a pilot booking centre about appointment statuses.

They asked whether we can add the 'Incomplete - other' status. At the
moment the incomplete appointments are broadly covered by ineligible age
and ineligible pension type.

However, there are other reasons why an appointment may be incomplete.

If we say ineligible age and ineligible pension count as incomplete at
the appointment stage it's sensible adding 'Incomplete - other'.


benlovell:
[Add 'Incomplete - Other' status to Planner](https://trello.com/c/4nbJu5VG/762-add-incomplete-other-status-to-planner)

PR #310, opened 2018-12-05
Add a notice to bookings with DC pot uncertainty

benlovell:
When the customer states they are unsure whether they have a DC pot, we
display a notice to the booking manager so they are aware this
booking/appointment may well be ineligible.


benlovell:
[Add visual prompt to manage appointment page for customers who are 'not sure' if they have a DC pension](https://trello.com/c/soPeJxSd/750-add-visual-prompt-to-manage-appointment-page-for-customers-who-are-not-sure-if-they-have-a-dc-pension)

PR #312, opened 2018-12-06
Reinstate the 'status' column

benlovell:
We've had several requests to reinstate the 'status' column. It makes
less sense to retain the 'requested' column as this is rarely considered
for appointments so we can remove this instead.


benlovell:
[Reinstate 'status' column](https://trello.com/c/1vOpHXd7/769-reinstate-status-column)

PR #313, opened 2018-12-12
Skip guider overlap validation at Reading

benlovell:
A specific Reading guider has 45 minute slots so we need to skip the
usual overlapping logic in this location, temporarily.


benlovell:
[Amend Planner validation rules for Reading location (45 minutes)](https://trello.com/c/F5J2B5tX/776-amend-planner-validation-rules-for-reading-location-45-minutes)

PR #316, opened 2018-12-19
Christmas notice period for emails

benlovell:


benlovell:
[As a customer I should be made aware of delays to the booking process over the Christmas 2018 period](https://trello.com/c/pI1jXwrs/731-as-a-customer-i-should-be-made-aware-of-delays-to-the-booking-process-over-the-christmas-2018-period)

PR #321, opened 2019-01-09
Allow booking managers to delete slots

benlovell:
This allows booking managers to delete realtime slots.


benlovell:
[Deleting slots using the 'realtime bookable slots' page](https://trello.com/c/n405XNFu/786-deleting-slots-using-the-realtime-bookable-slots-page)

PR #323, opened 2019-01-10
Overlapping appointments are not available

benlovell:
Prior to this change, a slot was still available when an appointment
overlaps but does not match the start precisely. This was causing
multiple, attempted bookings and contention for slots.


benlovell:
[Investigate duplicate booking requests at Coventry](https://trello.com/c/Mg3vOgAB/796-investigate-duplicate-booking-requests-at-coventry)

PR #324, opened 2019-01-10
Display all appointments on realtime calendar

benlovell:
Since the realtime calendar is per-day, it's highly unlikely that mixed
realtime/non-realtime appointments exist within a single view. Trying to
determine which appointments are realtime by overlaying them on realtime
slots is not entirely fool-proof and was causing inaccurate results.

This change simplifies the view and displays all appointments,
regardless.


benlovell:
[Investigate duplicate booking requests at Coventry](https://trello.com/c/Mg3vOgAB/796-investigate-duplicate-booking-requests-at-coventry)

PR #329, opened 2019-02-22
Display temporary email notice

benlovell:
We're having trouble delivering to `btinternet.com` addresses.


benlovell:
[Investigate possible email deliverability issues to BT internet addresses](https://trello.com/c/AEleD5Wg/907-investigate-possible-email-deliverability-issues-to-bt-internet-addresses)

PR #332, opened 2019-03-06
Populate organisation lookups by location ID

benlovell:
For each appointment location we need to map to the owning organisation
(CAS, CITA, NICAB) for a blazer report to analyse availability. Blazer
smart columns and variables don't give us what we need so the lookups
need to be local to the planner schema.


benlovell:
[As an analyst, I need to know how many bookable slots are free in TAP and Planner to ensure correct resourcing is in place](https://trello.com/c/Dsmpe4bK/871-as-an-analyst-i-need-to-know-how-many-bookable-slots-are-free-in-tap-and-planner-to-ensure-correct-resourcing-is-in-place)

PR #333, opened 2019-03-06
Avoid validation during redaction

benlovell:
This was causing issues with slot allocation and validation is not
necessary at this stage.


benlovell:
[Remove booking/customer details from Planner #7](https://trello.com/c/CnvxrZXG/916-remove-booking-customer-details-from-planner-7)

PR #334, opened 2019-03-15
Do not return full availability for locations

benlovell:
When new locations are added they will return full AM/PM availability
until a schedule is created. This change ensures new locations now
return no availability until a schedule is created.


benlovell:
[Remove the default full schedule behaviour on Planner when adding a new appointment location](https://trello.com/c/RdjXY1kg/860-remove-the-default-full-schedule-behaviour-on-planner-when-adding-a-new-appointment-location)

PR #335, opened 2019-03-22
Display appointment location

benlovell:
When the booking manager changes the location in the form they sometimes 
get confused as to what the customer-selected location was. This change 
ensures they will always understand which location the customer chose.


benlovell:
[As a booking manager, I want to see the location at which an appointment has been booked on the customer details screen.](https://trello.com/c/PGuPCFrw/743-as-a-booking-manager-i-want-to-see-the-location-at-which-an-appointment-has-been-booked-on-the-customer-details-screen)

PR #336, opened 2019-03-25
Appointments can be filtered by DC pot options

benlovell:
Allows the booking managers to filter and search appointments by the
customer's stated confirmation of their DC pension pot.


benlovell:
[Add 'DC Pension' filter to search bar](https://trello.com/c/bmhc2Z1q/787-add-dc-pension-filter-to-search-bar)

PR #338, opened 2019-03-28
Create default, blank schedules for locations

benlovell:
When a new location appears that would otherwise require a new schedule
to be created, we should automatically create a default, blank schedule
on the booking manager's behalf.

### DIFF
--- a/app/helpers/schedule_helper.rb
+++ b/app/helpers/schedule_helper.rb
@@ -42,8 +42,6 @@ module ScheduleHelper
   end
 
   def availability_button(location_id)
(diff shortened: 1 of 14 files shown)


benlovell:
[As an organisation admin on locator/planner, I need to be able to create a location and enable visibility, online booking and real time booking so that booking managers can set real time availability immediately](https://trello.com/c/WWD94Yn2/950-as-an-organisation-admin-on-locator-planner-i-need-to-be-able-to-create-a-location-and-enable-visibility-online-booking-and-real)

PR #339, opened 2019-04-03
Improve confirmation page for agents

benlovell:
Updated content and now mentions exactly which type of confirmation the 
customer should expect to receive.


benlovell:
[As a TP agent I require updated script on the F2F agent form so that I do not refer to am/pm slots when booking appointments on behalf of customers](https://trello.com/c/1Ouii4iU/779-as-a-tp-agent-i-require-updated-script-on-the-f2f-agent-form-so-that-i-do-not-refer-to-am-pm-slots-when-booking-appointments-on)

PR #340, opened 2019-04-12
Include guider name in email

benlovell:
Includes the guider name in the email so booking managers and guiders can
easily identity which booking and appointment notifications require their
attention.


benlovell:
[As a Pension Wise Booking Manager, I need the allocated guider name to appear on booking notification and cancellation emails so I can filter these easily](https://trello.com/c/lu1NXihZ/964-as-a-pension-wise-booking-manager-i-need-the-allocated-guider-name-to-appear-on-booking-notification-and-cancellation-emails-so)

PR #342, opened 2019-04-16
Restore the link to booking requests

benlovell:
Though these are now considered legacy due to realtime upgrades, there
are still cases where booking managers need to access legacy bookings
for historical reasons.


benlovell:
[Restore bookings link](https://trello.com/c/AxAkchrU/972-restore-bookings-link)

PR #345, opened 2019-05-17
Display timestamp in correct timezone

benlovell:
This wasn't respecting timezones previously.


benlovell:
[Correct times shown on Planner realtime slots page](https://trello.com/c/H6etT04D/994-correct-times-shown-on-planner-realtime-slots-page)

PR #348, opened 2019-05-29
Reduce grace period

benlovell:
Reduces the grace period in response to RTAP consultation and workshop.

### DIFF
--- a/app/lib/grace_period.rb
+++ b/app/lib/grace_period.rb
@@ -4,7 +4,7 @@ class GracePeriod
   end
 
   def start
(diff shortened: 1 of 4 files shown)


benlovell:
[As a face to face service we need to reduce the grace period by 1 working day so that customers have increased short term availability](https://trello.com/c/5ZlNFhsE/623-as-a-face-to-face-service-we-need-to-reduce-the-grace-period-by-1-working-day-so-that-customers-have-increased-short-term-availa)

PR #351, opened 2019-06-12
Don't permit a continual cycle of drop notifications

benlovell:
When a booking manager creates an appointment and the confirmation email to
themselves (in the role of customer) fails, they will be sent the booking
manager's regular email failure notification email. If this subsequently
fails, this enters an infinite loop of bounce/retry until the application
queue - in this instance, redis - falls over with out-of-memory exceptions.

This change ensures that whenever the booking's customer email matches any
of the found booking manager emails the drop processing is aborted and a
successful status is reported to mailgun, ensuring they do not continually
retry.

### DIFF
--- a/app/controllers/mail_gun/drops_controller.rb
+++ b/app/controllers/mail_gun/drops_controller.rb
@@ -3,6 +3,9 @@ module MailGun
     skip_before_action :verify_authenticity_token
 
     def create
+      form = DropForm.new(drop_params)
+      form.create_activity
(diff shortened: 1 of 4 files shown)


benlovell:
[Issue with mail delivery notifications and Planner (website and agent booking)](https://trello.com/c/0OgjESJQ/1017-issue-with-mail-delivery-notifications-and-planner-website-and-agent-booking)

PR #355, opened 2019-06-28
Send final reminder at 72 hour interval

benlovell:
We previously sent reminders at 48 hours before the appointment was due
to take place. This has been extended to 72 hours in the hope that
cancellations grant enough time for their underlying slots to be
rebooked by others in the current booking window.


benlovell:
[Change 48hr sms and email reminder to 72hrs](https://trello.com/c/ZxKfxJwU/1040-change-48hr-sms-reminder-to-72hrs)

PR #356, opened 2019-07-03
Remove multiple choice slots from agent journey

benlovell:
These are now unnecessary since we are fully realtime across all
locations.


benlovell:
[Update Planner agent form](https://trello.com/c/7rPkXHiH/1048-update-planner-agent-form)

PR #357, opened 2019-07-11
Allow TP agents to see all booking

benlovell:
Previously they could only view bookings made by agents.s


benlovell:
[As a TP Agent Manager, I need to see all F2F appointment bookings so that I can respond to customer queries / complaints and action requests received via Zendesk](https://trello.com/c/gMxQephK/937-as-a-tp-agent-manager-i-need-to-see-all-f2f-appointment-bookings-so-that-i-can-respond-to-customer-queries-complaints-and-action)

PR #358, opened 2019-07-11
Only send to actual booking managers

benlovell:
This was highlighted by an edge case whereby a re-synchronised and
previously inactive user was notified for a new booking although they
were no longer assigned the correct permissions for this organisation
and application.


benlovell:
[As a sign on user I should not receive f2f booking notifications unless I have booking manager permissions with a specific organisation](https://trello.com/c/1hWvOHRl/1061-as-a-sign-on-user-i-should-not-receive-f2f-booking-notifications-unless-i-have-booking-manager-permissions-with-a-specific-organ)

PR #360, opened 2019-07-17
Ensure only booking managers are notified

benlovell:
These places were missed last time around. This ensures only booking
managers will receive these particular notifications.


benlovell:
[As a sign on user I should not receive f2f booking notifications unless I have booking manager permissions with a specific organisation](https://trello.com/c/1hWvOHRl/1061-as-a-sign-on-user-i-should-not-receive-f2f-booking-notifications-unless-i-have-booking-manager-permissions-with-a-specific-organ)

PR #361, opened 2019-07-18
Drop labelling for agent slots

benlovell:
This was missed before.


benlovell:
[Update Planner agent form](https://trello.com/c/7rPkXHiH/1048-update-planner-agent-form)

PR #363, opened 2019-07-19
Notify customers upon cancellation

benlovell:
Notifies the customer by email when their appointment was cancelled by a
booking manager.


benlovell:
[As a F2F customer I need a cancellation email so that I know my appointment has been cancelled](https://trello.com/c/XyjJge2Y/603-as-a-f2f-customer-i-need-a-cancellation-email-so-that-i-know-my-appointment-has-been-cancelled)

PR #366, opened 2019-08-01
Allow agents to edit bookings

benlovell:
Gives agents the ability to edit particular fields on appointments. They 
can also leave activity entries and re-send email confirmations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ addons:
   chrome: stable
 services:
 - redis-server
+- postgresql
 cache:
   bundler: true
   directories:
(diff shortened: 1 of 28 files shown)


benlovell:
[As a TP Agent Manager, I need to be able to update basic customer details and resend confirmation emails so that I can action customer requests received via Zendesk](https://trello.com/c/LXbSGSNA/1062-as-a-tp-agent-manager-i-need-to-be-able-to-update-basic-customer-details-and-resend-confirmation-emails-so-that-i-can-action-cus)

PR #367, opened 2019-08-14
Log overlapping reasons for instrumentation

benlovell:


benlovell:
[Expand logs to identify the overlapping schedule when real time appointments are moved and an underlying slot not created](https://trello.com/c/APiJzUsm/1089-expand-logs-to-identify-the-overlapping-schedule-when-real-time-appointments-are-moved-and-an-underlying-slot-not-created)

PR #368, opened 2019-08-23
Aliasing for customer name search

benlovell:


benlovell:
[Include table alias when searching by customer name](https://trello.com/c/lnc7eUVc/1103-include-table-alias-when-searching-by-customer-name)

PR #369, opened 2019-08-28
Ensure agent managers can resend emails

benlovell:


benlovell:
[Ensure agent managers can resend email confirmations](https://trello.com/c/t5eX3MSt/1104-ensure-agent-managers-can-resend-email-confirmations)

PR #370, opened 2019-08-29
Remove old-style slots

benlovell:
Addresses the issues with slots and formatting. Prior to this we were using
combination of date/start/end for bookable slots and this is complicated
for many reasons. Moving towards a simple date/time for
`start_at` and `end_at` removes most of the complexity.

This was also a good time to remove redundant old-style slot features.


benlovell:
[Refactor how dates and times are processed on Planner](https://trello.com/c/hznqZiGz/1078-refactor-how-dates-and-times-are-processed-on-planner)

PR #373, opened 2019-09-30
Validate additional info requirement

benlovell:
When accessibility requirements are specified we need to ensure the
customer/agent supplies descriptive text for their needs to be
fulfilled.


benlovell:
[Changes to accessibility checkbox and wording](https://trello.com/c/RYLU34ju/1137-changes-to-accessibility-checkbox-and-wording)

PR #374, opened 2019-10-16
Improve pension provider labelling

benlovell:
Agents are still incorrectly completing this field so improving the
label should help make things clearer.


benlovell:
[Amend wording for final nudge drop down (agent booking forms)](https://trello.com/c/PFcwk7yF/1161-amend-wording-for-final-nudge-drop-down-agent-booking-forms)

PR #375, opened 2019-10-23
Ignore webhooks for non-customer emails

benlovell:
We ought to ignore webhook calls for message types not referring to customer emails. Drop notifications for booking managers can get tied to bookings in some instances, particularly when the booking email is the same as a current booking manager. Only caring about customer emails  ensures this never happens.

### DIFF
--- a/app/forms/drop_form.rb
+++ b/app/forms/drop_form.rb
@@ -5,6 +5,14 @@ TokenVerificationFailure = Class.new(StandardError)
 class DropForm
   include ActiveModel::Model
 
+  INTERESTING_MESSAGE_TYPES = %w(
+    appointment_modified
+    appointment_confirmation
+    appointment_reminder
+    appointment_cancellation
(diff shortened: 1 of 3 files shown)


benlovell:
[Investigate why Mailgun has repeatedly attempted to send appointment confirmation email months after the scheduled appointment was cancelled](https://trello.com/c/5jyE9xXh/1168-investigate-why-mailgun-has-repeatedly-attempted-to-send-appointment-confirmation-email-months-after-the-scheduled-appointment-w)

benlovell:
[Investigate why Mailgun has repeatedly attempted to send appointment confirmation email months after the scheduled appointment was cancelled](https://trello.com/c/5jyE9xXh/1168-investigate-why-mailgun-has-repeatedly-attempted-to-send-appointment-confirmation-email-months-after-the-scheduled-appointment-w)

PR #376, opened 2019-10-25
Exclude new year holidays

benlovell:
Adds exclusions for CAS and others.


benlovell:
[Christmas & New Year Closure Dates 2019/20](https://trello.com/c/33r6eUyL/1171-christmas-new-year-closure-dates-2019-20)

PR #380, opened 2019-11-11
We send the final SMS reminder at 72hrs

benlovell:


benlovell:
[Change Planner / TAP agent script so it refers to 72 hours for sms reminders](https://trello.com/c/C5lIvtLH/1190-change-planner-tap-agent-script-so-it-refers-to-72-hours-for-sms-reminders)

PR #384, opened 2019-12-11
Configure Mailgun validation key

benlovell:


benlovell:
[Inline email validation via Mailgun is failing](https://trello.com/c/Vqw6ltP4/1217-inline-email-validation-via-mailgun-is-failing)

PR #386, opened 2019-12-17
Remove inline mailgun validation

benlovell:
It's too unreliable and clearly not production-ready.


benlovell:
[Inline email validation via Mailgun is failing](https://trello.com/c/Vqw6ltP4/1217-inline-email-validation-via-mailgun-is-failing)

PR #391, opened 2020-01-14
Add holidays across all organisations

benlovell:
Adds holidays across all three organisations. This relies on the
organisation lookups being current so I've scheduled a nightly job to
reimport them. This will pick up changes made to locations and their
children.


benlovell:
[As a booking manager, I need the 2020 Bank Holidays added to Planner as mandatory closure days](https://trello.com/c/D743YaI5/1245-as-a-booking-manager-i-need-the-2020-bank-holidays-added-to-planner-as-mandatory-closure-days)

PR #392, opened 2020-01-16
Display potential duplicates

benlovell:
Show appointments from the same booking location that share either the
customer's name, email or phone.


benlovell:
[Highlight duplicate bookings which exist across a booking centre](https://trello.com/c/IIon6u15/919-highlight-duplicate-bookings-which-exist-across-a-booking-centre)

PR #394, opened 2020-01-24
Include access information in confirmation emails

benlovell:
Includes access information - when specified - in the customer emails.


benlovell:
[As a F2F customer I want venue access information available on the Pension Wise website and automated emails so that I can access my appointment](https://trello.com/c/zE51smxr/904-as-a-f2f-customer-i-want-venue-access-information-available-on-the-pension-wise-website-and-automated-emails-so-that-i-can-acces)

PR #396, opened 2020-01-31
Allow admins to clear future slots

benlovell:
Administrators can clear future slots from a schedule. This is handy
when a location with scheduled slots goes inactive. In these cases the
guiders will be reallocated to other locations but when they exist in
overlapping slots elsewhere this is not possible.

### DIFF
--- a/app/controllers/realtime_bookable_slots_controller.rb
+++ b/app/controllers/realtime_bookable_slots_controller.rb
@@ -38,6 +38,16 @@ class RealtimeBookableSlotsController < ApplicationController
     end
   end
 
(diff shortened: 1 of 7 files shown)


benlovell:
[As a Sign On administrator, I need the ability to delete all future slots in a Planner schedule so that I can reduce overlapping slot errors](https://trello.com/c/CcrH7lEL/1278-as-a-sign-on-administrator-i-need-the-ability-to-delete-all-future-slots-in-a-planner-schedule-so-that-i-can-reduce-overlapping)

PR #399, opened 2020-02-27
Remove DC pot question for agents

benlovell:
They should be establishing this up front now so no need to ask the
question explicitly. We can default it to 'yes'.


benlovell:
[Remove defined contribution pot confirmed question from Planner](https://trello.com/c/CEbDok0K/1316-remove-defined-contribution-pot-confirmed-question-from-planner)

PR #405, opened 2020-03-05
Remove final nudge options

benlovell:
The underlying data remains but the API does not log the provider
neither does the agent booking form permit these values.


benlovell:
[Remove final nudge urls, booking journey messaging and TP agent outcome form & link](https://trello.com/c/BG3aTSU8/1306-remove-final-nudge-urls-booking-journey-messaging-and-tp-agent-outcome-form-link)

PR #407, opened 2020-03-19
Signpost to self-serve for cancellations

benlovell:


benlovell:
[As a customer I should be signposted to the explore your options pages via cancellation email so that I can find out more about my pensions options](https://trello.com/c/re8CGxpL/1348-as-a-customer-i-should-be-signposted-to-the-explore-your-options-pages-via-cancellation-email-so-that-i-can-find-out-more-about)

PR #482, opened 2023-08-01
upgrade to ruby 2.7.7

eyss1234:


philManWithBeard (commented):
Is ruby 2.7 still supported?

benlovell:
> Is ruby 2.7 still supported?

In theory any 2.7 series ruby should still be fine.

